PR #3, opened 2016-12-05
Adding a way to print span/trace ids

lookfwd:
Adding a method that prints minimal information able to identify the span (e.g. trace id/span id) to an output stream. This is useful in order to be able to correlate traces with other types of logs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,12 @@ make
 ./test
 ```
 
+Note - You might have to set your `LD_LIBRARY_PATH`:
+
+```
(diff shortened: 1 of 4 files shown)


bensigelman:
This is something that's come up a number of times... I just summarized the issue here: https://github.com/opentracing/specification/issues/24

tedsuo:
@lookfwd do you mind if I close this PR?

lookfwd:
no, bo problem!

PR #5, opened 2016-12-23
keep gcc specific flags specific to a gcc build

SaintDubious:
Adds a COMPILER_IS_GCC flag in the configure.ac, which can then be read in Makefile.am. Then restrict the gcc specific flags to when gcc is the compiler. With this in place we can build on AIX and Solaris


lookfwd:
LGTM. @jmacd, @bensigelman  ?

PR #6, opened 2017-01-11
Opentracing-cpp Redux

jquinn47:
This is a significant rewrite of the current opentracing-cpp implementation. I've been working on this with @lookfwd  and @SaintDubious internally for the last couple weeks.

### Major Changes

I took issue with the original implementation in a few significant ways:

#### Library Organization

The original implementation required tracers to `dynamic_cast` `Writers` and `Readers` into a concrete type before they could be used. This is an anti-pattern.

We would go down one of two roads.

1) All of the Carriers are defined in the Implementation library

* This makes it very difficult for organizations to pull down an opensource (e.g, zipkin) implementation and add their own carriers to it. If I have a `OrganziationTransportRPC`, there is no way that's getting pushed back into the open source implementation.

* If an organization writes their own implementation, the new library needs to be dependent on _every_ RPC mechanism in the organization that may need to use the tracing library. Obvious side effects with code size and dependency management.

2) All of Carriers are defined external to the implementation library

* If the Tracer does not have the writers defined in the library, we've created a physical circular dependency between libraries.  (Tracer needs to know about Carriers, Carriers need to know about Reader/Writer defined next to Tracer). Dependency hell ensues

_Heavy use of CRTP templates to define Reader/Writer interfaces. Functors can be passed into the API at compile time without affecting the base library. Implementations have an interface to rely on for more OO approach_

#### Allocations
The original implementation always used `new`/`delete` and exposed raw pointers to calling code.

Without an interface to pass pointers back to the Tracer, this removes the ability to use custom allocators or object pools for managing spans/span contexts. 

_Added the `Tracer::cleanup()` method for returning resources to the Implementation. Introduced a `StringRef` to avoid unnecessary creation of temporary `std::strings` at the interface level_

#### V-Table hits
In high-throughput scenarios, the v-table is not desirable. If we're planning on providing an API for performant C++ applications, we would ideally be able to avoid as much overhead as possible in the applications we're instrumenting. 

_We've worked around this with a compile time pattern for polymorphism._

#### Spec concerns
We were missing a few things in the spec:
* constants
* being able to iterate through baggage
* binary writer

### Minor Changes
* Added lots of documentation/usage examples
* Shifted to the cmake build system for better integration with gtest
* Primary library is header only for easier installs (Noop requires the .a be built, but its optional)
* Applied some clang-format tools to keep the code base consistent (I just used one my team uses internally, but I'm fine if we want to change the format so long as its consistent 👍  )

This is obviously a large change and warrants a major version bump. We've kept @lookfwd  in the loop the whole time, as he's definitely our most active opentracing advocate. I believe he's on board but I'd want him to chime in here too.


bensigelman:
I had a chance to look at this. It's really a huge shame to do this much detail-work without even checking in with all interested parties in advance... a lot of decisions (e.g., whether CRTP is a pattern that OpenTracing would like to embrace, whether a vtable lookup is literally ever relevant from a cost standpoint when the things being measured take ≥ 1ms, whether to prioritize the experience of the programmer *using* the OpenTracing API or the experience of the programmer *implementing* the OpenTracing API, etc, etc) could and definitely should have been discussed before any non-prototype code was written.

As-is, my sense is that #6 is a clean, well-tested, well-documented, performant implementation of an API that the various stakeholders have not actually agreed upon. We are now in the unenviable position of trying to come to that agreement with the code already written.

How to proceed: I would like to return to where this conversation probably should have started... what are the actual priorities here? In my mind, they are as follows:
- Should this target `C++11`? Or `C++98`? or `C`?
- The API must be clean and easy for a *caller* to understand (even if that means that implementors have to do things we don't adore as C++ programmers)
- Expected performance must be appropriate to measure Spans with expected durations ≥ 100µs... anything faster than that is a weird fit for distributed tracing where end-to-end distributed durations are typically milliseconds or longer.

lookfwd:
> It's really a huge shame to do this much detail-work without even checking in with all interested parties in advance.

No pressure. This code would be written anyway, I think. P.S. I think the "interested parties" have been very few and quite busy lately.

> Should this target C++11? Or C++98? or C?

This has been discussed a lot in the past. First page of README.md; C++98

> ≥ ~1ms,  ≥ 100µs... 

This is also my typical experience, but if there are such numbers in mind or used as design constraints, they should be explicit. It might help people decide early if OT is relevant to them

That said, I don't think that the drive for this is (just) performance, v-table lookups and other such things. The existing API is a nice toy API but as soon as we got it through some more people ... they just couldn't integrate it into their codebase. It's full of holes on the semantics and also quite out of date with the spec itself (as it evolved in the last 6 months).

Did you @bensigelman or anyone else in the OT community try to sell the existing C++ API to anyone and can share some experiences? Because this PR is the result of trying to do so. People wouldn't accept the existing API. It was polluting their codebase with bad code from the instrumentor's standpoint. This is an MVP that is expected to play well with some serious code i.e. with code that will last for 5+ years and will go to 1000+ services.

jquinn47:
I'll just echo @lookfwd's points. There were many problems with the initial implementation that I've outlined in my initial comments. I did make an attempt to reach out in the gitter, but the only response I got was from @lookfwd at the time. With his recommendation, we had moved forward. Apologies if we caught you off guard. 

With that said, I'm not opposed to changes, but I believe we've already addressed the points you've brought up:

* As for the the C v C++98 v C++03 v C++11 v C++XY, I tried to stick with the projects initially stated goals of C++98. I believe I missed the mark with regards to the use of `size_t` and `uint64_t` as they were brought in C99. This has simple workarounds with `cmake` testing for `size_t`/`stdint.h`. Otherwise, I don't believe I've rocked the boat here.

* The semantics of the interface are very similar to the `opentracing-go` API. I've outlined how that would work here: https://github.com/jquinn47/opentracing-cpp/blob/master/docs/usage.md#creating-spans

* I can't speak much to the performance of an implementation here, as there is no implementation to measure. For the applications which are performance sensitive, it would be ideal to avoid unnecessary memory allocations, or hit the v-table. 

* I would not disregard the current implementations library organization problems. Avoiding a physical circular dependency (or requiring implementations link in every other RPC library) should be first class citizens in the discussion

lookfwd:
If it helps, let's organize a Google Hangout with everyone interested.

bensigelman:
@lookfwd is right that we should schedule a hangout to discuss.

First off: nobody is defending or arguing for what's presently in opentracing-cpp: it is truly just a toy and probably shouldn't be in master.

The major interested parties are probably people concerned with Lyft's Envoy project... of course several of them are at Lyft, but there are also folks who want to see OpenTracing-cpp happen from IBM, Google, and I think also Yahoo-Japan? Though this is all from memory. Not all of this interest is visible on Gitter. (PS: sorry I missed the note on Gitter reaching out about this in the first place)

Regarding other implementations: there is a LightStep C++ instrumentation library that is very much OpenTracing-inspired, and which was always intended to inform possible designs for opentracing-cpp... we have used it to work out C++ API details in the context of a production codebase (rather than in isolation). For instance:

https://github.com/lightstep/lightstep-tracer-cpp/blob/master/src/c%2B%2B11/lightstep/tracer.h

As you will see, the API targets C++11, is quite different than the proposal here, etc.

Anyway, getting back to @lookfwd's suggestion: can we schedule a time for a hangout to talk through some of the higher-order bits? Thanks...

lookfwd:
I will refer to [gitter](https://gitter.im/opentracing/opentracing-cpp) for the C++98 discussion. I spend hours to get agreement on that before the first release of the current version, on June. I can't make Lightstep adhere to OT though... it's your API.

lookfwd:
I like the Envoy code and their CI system very much. Awesome "out of box" experience, 700 tests + heap profiler, fantastic! In terms of tracing they essentially have their own wrappers and they've limited lightstep in a main class [here](https://github.com/lyft/envoy/blob/009d341ac8c6d1f70ceaa725b907086dc20587d9/source/common/tracing/http_tracer_impl.h#L102) and a less critical [here](https://github.com/lyft/envoy/blob/009d341ac8c6d1f70ceaa725b907086dc20587d9/source/common/tracing/http_tracer_impl.h#L143). It should be reasonably easy to plug OT. The only thing that seems quite non-OT standard is that I can't find them using (lightstep's) `Inject` and `Extract`, effectively bypassing these more challenging parts of the API. I suspect that Envoy creates the `RequestID` [here](https://github.com/lyft/envoy/blob/4332dcf5c47dd93ce070166830884223515b3b67/source/common/http/conn_manager_utility.cc#L112) (and propagates it [here](https://github.com/lyft/envoy/blob/4332dcf5c47dd93ce070166830884223515b3b67/source/common/http/conn_manager_utility.cc#L137) and [here](https://github.com/lyft/envoy/blob/009d341ac8c6d1f70ceaa725b907086dc20587d9/source/common/ratelimit/ratelimit_impl.cc#L50)) which is later used as trace - request id (`guid:x-request-id`) [here](https://github.com/lyft/envoy/blob/009d341ac8c6d1f70ceaa725b907086dc20587d9/source/common/tracing/http_tracer_impl.cc#L249). Overall, from the code structure, the async-api, the thread locals and some comments relating to eliminating string copies, I get the feeling that Envoy developers are very serious about performance and this means that likely we're all in the same page.

lookfwd:
Correction - there's been some more OT friendly recent work (last 10 days) in [another branch](https://github.com/lyft/envoy/commits/refactor_for_parent_child/include/envoy/tracing/http_tracer.h). This explains a few things...

lookfwd:
@mattklein123 @RomanDzhabarov - Hello!

jquinn47:
From what I gathered we're all in different timezones. US EST here, but I'm available for a hangout whenever is doable. I let the folks with more extreme differences pick the time 👍 

In the mean time, I'll have to go through lyft/lightstep as @lookfwd did to get a feel for whats being attempted elsewhere

SaintDubious:
I think having a real time conversation would be helpful at this point. I apologize that such a large change was dumped in your lap so abruptly. We (Jim, Dimitrios and myself) have a pressing need to get a working implementation of Trace online and we had serious concerns about the existing API, as Jim has outlined. As far as we could tell there wasn't an active community around the C++ API (the gitter channel and github C++ repos were quiet) so we felt free to take our best shot at creating a better API.

I am very impressed with Jim's work. While I grant that it is not as easy to understand as a more traditional inheritance model, we all feel it is a much more solid approach. I also want to point out that he didn't go off the rails and do anything crazy for this, his use of static polymorphism, adapters, and the CRTP are all accepted patterns that are well understood and documented in the C++ community.

I'd be very interested to get feedback from some of the other potential users of this that @bensigelman mentioned. While it may not be the perfect API, I think we can all agree that it's better then what exists and I'd like to see it made available to the entire community.

mattklein123:
Hi,

A few high level comments about this:
- Unfortunately I don't have the time to go through this change in any detail. @RomanDzhabarov should have a bit more time to look through the API. Just from reading the description, avoiding things like virtual dispatch seems like over-optimization to me and would give me pause, but as long as the code works and satisfies the following bullet points I don't think we will have any problem consuming it.
- We are planning on using inject/extract soon to create parent/child span relationships. The only reason this isn't done now is getting things out iteratively has been a priority, so we should consider this moving forward from the Envoy perspective.
- Our main objective WRT tracing in the near term for Envoy is to have Zipkin support. Beyond the companies that @bensigelman already listed above, we have a few other large organizations that want Zipkin support in Envoy (I haven't checked in with them recently about publicity so I can't name them here). From our viewpoint, the best way of getting Zipkin support is to move Envoy to the OT C++ API, and then have "drivers" for both LightStep as well as Zipkin.
- The main thing we care about in Envoy is that a library like OT C++ should separate transport from serialization. We want (require) that we do all of our own transport for performance/threading reasons, so the API should allow the decoupling of those two parts of OT. This is how we are currently doing it in the LightStep library and we would like to keep that for OT C++.
- It is our large desire that the OT C++ should mostly be a drop in replacement within the small number of files in Envoy that you already pointed out. So instead of the LightStepSink and LightStepRecorder we will be using OpenTracingSink and OpenTracingRecorder (approximately), details have to be worked out in terms of how the "driver" would perform serialization and then allow Envoy to do transport.
- Within the confines above, though I would prefer we just use C++11, I don't really care if it's C++98, C, etc. We will consume/wrap whatever is produced and working as long as it satisfies the above.

@RomanDzhabarov should be your main contact for the Envoy side of this.

cc @adriancole @rshriram

lookfwd:
@mattklein123 - everything you say makes sense, and the approach with the sink etc. and the C++-zipkin interface approach etc.

> The main thing we care about in Envoy is that a library like OT C++ should separate transport from serialization.

With this you mean is out-off-band data i.e. what you now do with `LightStepRecorder` and you send data to LightStep. As far as I'm aware, OT doesn't specify this part and it's your contract with a tracer implementation, the way you use to send data to the infrastructure. Please correct me if I'm  wrong. This would mean that you need no changes to support OT as long as (the existing factory) is passed on the constructor as you do right now [here](https://github.com/lyft/envoy/blob/009d341ac8c6d1f70ceaa725b907086dc20587d9/source/common/tracing/http_tracer_impl.cc#L213). An unusual thing around is that you have one tracer per thread if I understand correctly. I would expect the thread local-ness to be hidden in the tracer and in general to have a single instance of a tracer. Not a big deal though.

> should separate transport from serialization.

Back to this topic, there's also in-band data that a tracer might require you to receive and send. For this, `inject`/`extract` can give you (or expect the system to provide them) a few tracer-specific headers. For in-band data, also, OT doesn't define anything about transport. Tracers do the serialization and with an efficient API it's all fast and nice / 0-copy'ish.

lookfwd:
I'm available for a hangout .. the usual US-friendly-insomniac-Dimitris hours i.e. around 18:00 - 20:00 EST  :)

mattklein123:
> As far as I'm aware, OT doesn't specify this part and it's your contract with a tracer implementation, the way you use to send data to the infrastructure

I'm not familiar with OT. If this is the case, that's perfect. The Envoy "drivers" can handle serialization/transport aspects the way that Envoy requires.

> An unusual thing is that you have one tracer per thread if I understand correctly

Not quite. There is a global tracer. The implementation hands the trace requests off to thread local connection management system, so it is in fact hidden inside of the "driver." In the new world, we would keep this model since thread local behavior is only needed for transport, not serialization.

> Back to this topic, there's also in-band data that a tracer might require you to receive and send

Yup we are aware of this. We have been working out a plan with @jmacd to base64 encode the provided information into a header that we will propagate through the mesh. This part isn't an issue. Roman is working on a refactor currently to set us up for this.

jquinn47:
C++11 is a deal breaker on our end unfortunately. We've got older legacy servers that just don't have that support yet. This would be typical of some of the larger, slower to move organizations, but we would all completely understand if C++11 is the decision for the `opentracing-cpp` community to make.

The predominant problem with all of the implementations I have seen so far is the requirement that they know about all of the carrier formats in advance. If we want to meet [this inject specification](https://github.com/opentracing/specification/blob/master/specification.md#inject-a-spancontext-into-a-carrier), you’re forced to write a C++ tracer that implements a dynamic switch on type (enum or cast). 

_This is an anti-pattern._ 

The tracer needs to know every possible carrier it’s going to interact with. If the carriers are all defined in the tracer library, the tracer library becomes monolithic. If the tracer does not define the carriers, but instead relies on other libraries to define them, we introduce a cyclic dependency between the tracer and those carrier libraries.

@SaintDubious and I spent the last week of December trying to work the problems, and the CRTP solution was what we came up with. The v-table removal optimization was an ancillary benefit, but not the primary reason for the design decisions.

adriancole:
>
> The predominant problem with all of the implementations I have seen so far
> is the requirement that they know about all of the carrier formats in
> advance. If we want to meet this inject specification
> <https://github.com/opentracing/specification/blob/master/specification.md#inject-a-spancontext-into-a-carrier>,
> you’re forced to write a C++ tracer that implements a dynamic switch on
> type (enum or cast).
>
> *This is an anti-pattern.*
>
> The tracer needs to know every possible carrier it’s going to interact
> with. If the carriers are all defined in the tracer library, the tracer
> library becomes monolithic. If the tracer does not define the carriers, but
> instead relies on other libraries to define them, we introduce a cyclic
> dependency between the tracer and those carrier libraries.
>
In practice, I think many tracers do not support open-ended carriers (ex
via registration etc). I suspect you can constrain to http (or make a base
variant that isn't open-ended), if that's a problem.

lookfwd:
> C++11 is a deal breaker on our end unfortunately.

I agree but also add that nobody prevents and I would be very happy if there's an (optional) well implemented C++11 Adapter/Layer in OT on top of C++98. I can imagine it to add 0 overhead. Additionally, nobody prevents Tracers to be implemented in C++11 as long as they fall-back to C++98 on the API boundary. Of course we can't guarantee that such tracers will be able to compile/link with C++98 binaries but that's no worse to the alternative. If one is convinced that their audience is C++11 - go for it! OT just needs to not prevent a C++98 implementation which unfortunately is sometimes the only option for "Big Iron" and some more embedded stuff.

A quick note on C; if there's an audience... I believe their needs are widely different to the >= C++98 audience. It's not a matter of just turning objects to struct's and having long function names. C means macros, pointer arithmetic and `strncmp`'s, `void*` and `malloc`s. I'm exaggerating but my feeling is that, that audience has severely different requirements.

@jquinn47 's argument on carriers/writters/readers etc. is unfortunately true as far as I can tell. It took me 2-3 days attempting to write something that solves this problem before I fully comprehend it.

In terms of supporting just HTTP or something more limited, my view is that at this relatively early stage an API should be able to handle whatever is in the Spec comfortably. If it has to stretch or start compromising (within the first year of its life) what are the chances for it to survive for a few years (without breaking backwards compatibility too much - which I think everyone in C++ hates).

In this specific case, I think the aim is to use some pretty standard IPC mechanisms (not HTTP) without having to do too many (string) copies. It's not about latency per se - it's about reducing the working set, scaling to larger concurrencies and having an API that doesn't require hacks (so early in its life).

jmacd:
This thread has become more of an issue discussion than a code review. I'll briefly add:

1. @jquinn47's comments about the OT-spec'd anti-pattern resonate with me, and there was a similar discussion for Golang: https://github.com/opentracing/opentracing-go/issues/127
In that issue discussion, I proposed introducing a new `CarrierInjector` interface to (at least partially) address this anti-pattern: Inject would take a single carrier/injector argument that would receive the (abstract) tracer, allowing us to write carrier types that dispatch to other carrier types, so that all Tracers do not need to know about all carrier types.

2. I believe C++11 is enough different from C++98 that the two deserve (at least consideration of) separate OT APIs. Since this conversation is being led by the effort to integrate Envoy and Zipkin we should at least realize that Envoy, because it is a framework for executing non-C++ code, is not an "every day" C++ tracing kind of application. When I put together the LightStep C++11 tracing API, the idea was for it to be easy to use in a C++11 code base, for me that means that span contexts (including baggage) can be propagated easily, memory management and lifetime management should be straightforward, without sacrificing efficiency. I think this PR shows that you *can do those things in C++98*, but I'm not sure you would do those things if you have C++11 at your disposal.

jquinn47:
Since the entire implementation is deferred, implementations are allowed to use C++11 regardless. It's only at the interface level where a C++11 interface can't leak C++11isms to C++98 clients. With that fact, I don't see a reason why we can't support both with a compilation switch. If that's agreeable, I'd add `ifdef OPENTRACING_CPP11` turning on the following:

In C++11 mode:
* Declare versions of tag/log/setBaggage acceptring rvalues for move semantics
* Returning std::unique_ptrs instead of raw pointers when creating spans/contexts
* Returning std::shared_ptr for `Tracer::instance()` to avoid lifetime issues on tracer swaps

In C++98, you get what we have here today. Raw pointers whose lifetime must be explicitly managed. This makes would make it difficult for current C++98 users to upgrade to OT-C++11, when/if they move to C++11, but I'm not ready to tackle that problem too.

This would be straight forward, however, if the complexity of the patterns I'm using is already a root issue, this would only make it more complicated, not less.

SaintDubious:
I'd like to curtail the C++11 conversation for this PR if at all possible. As @lookfwd pointed out, the Readme.md states C++98, and unfortunately our hands are tied, we require C++98. 

Thank you, @jmacd, for the link to the related conversation in Go. It's very interesting to see that this problem exists in other languages. @jquinn47 and I spent a long time trying to get rid of the "switch on type" anti-pattern, but ultimately decided that dynamic polymorphism just won't work.  We felt that a good solution would let the compiler select the correct Carrier, and then tried a number of approaches to make that happen. We ended up on the [Curiously Recurring Template Pattern (CRTP)](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) because it checked all the boxes and it is well understood. 

As @jquinn47 continues his work on the Bloomberg specific Tracer implementation of this API I am happy with how the CRTP is working:
- Usage of his new API for an end user is straight forward
- Interaction with it is through well defined interfaces (GenericSpan et all) which should allow for easy replacement with another implementation 
- The code for inject and extract is short and obvious, with the compiler selecting or generating the code for different Carriers.

In short, this PR proposes an API that works very well. I think we've discussed the points raised by @bensigelman in his initial response. I'd like to do what I can to help get this PR merged. Are there any specific concerns with what @jquinn47 has written? Perhaps some Tracing implementers would like to discuss our experience with it so far? @jquinn47 and I are available.

docs/usage.md
```diff
@@ -0,0 +1,323 @@
...
+
+### Outline
+
+The `opentracing-cpp` interface uses the [Curiously Recurring Template Pattern](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern)
+(CRTP) to define a set of compile time, polymorphic interfaces to `OpenTracing` implementations.
+
+In order to trace their systems, clients must:
+
+ - Select their `opentracing-cpp` implementation
+ - Install the selected `Tracer` in their applications in main
```
yurishkuro: Must this be a MUST? There are cases when we want to use different instances of the tracer in the same application.

docs/usage.md
```diff
@@ -0,0 +1,323 @@
...
+    return 0;
+}
+```
+
+### Defining carriers
+
+In order to instrument our entire network, we need a way to `inject` and `extract`
+spans into and out of our RPC calls. Extending our example, when we make our request
+to the backend HTTP service, we'll need a way to inject the details of the context
+into our outgoing HTTP request.
```
yurishkuro: I may be missing something, but how is the specificity of HTTP protocol, which does not guarantee preservation of the header names, addressed (cf. [this comment](https://github.com/opentracing/specification/issues/15#issuecomment-273447780))?  The generic text-map injectors are not suitable for HTTP.

docs/usage.md
jquinn47: I assumed there was only one, global tracer. My original implementation had a `Tracer::instance()` method that returned a pointer, but seeing `Tracer::instance()` for every call seemed redundant, so I wrapped the static methods around those calls.

This static interface is important for library developers.

The library developers don't have access to main, nor do they know about custom tracers being installed. The single point of entry to the `Tracer` removes that problem.

We could have an interface to return multiple tracers, but then all callers at all call sites then need to loop over instances. @lookfwd  had an elegant idea of higher order tracers, which we could install more primitive tracers into the top level global tracer. This would be particularly useful for transition from one implementation to another, but during the transition period, using both.

With all that being said, there's nothing stopping multiple instances of TracerImpl being stored independent or instead of the `instance()` method; you lose the singleton benefits, but gain multiple instances if thats a requirement.

docs/usage.md
jquinn47: HTTP wasn't handled at all. I started this when TextMap and Binary were the only protocols. I'd add `GenericHttpReader`/`GenericHttpWriter` for those use cases, but haven't gotten to it yet.

This usage example is likely misleading for the reasons you bring up.

jquinn47: I'd want to see the outcome of that HTTP issue before I started to implement the HTTP carriers, but the usage example could be cleaned up regardless.

bensigelman:
@jquinn47 @SaintDubious @lookfwd 

Can you propose a day or two that would work for a 6pm ET or 7pm ET hangout to discuss this PR? We are pretty well aligned on higher-level goals but still have too many things to resolve for a single PR. I think an actual 1h conversation will be enough to make substantial progress on the remaining (major) sticking points and we can go from there to address the usual pedantic code-review stuff. :)

Thanks.

lookfwd:
I would go for e.g. coming Tuesday. If we do 30 mins (up to 1h) and potentially repeat it a few days later if necessary we might be more effective :)

SaintDubious:
Sounds like the best way forward. @jquinn47 and I are available Tuesday as well. In the meantime, if it's possible to enumerate the remaining sticking points that might help us in preparing more thoughtful responses.

bensigelman:
@SaintDubious @jquinn47 @lookfwd @mattklein123 @yurishkuro @jmacd let's try for next Tuesday (Jan 24) at 6pm ET / 3pm PT [since @lookfwd said that would work for him]. I blocked an hour but of course we will end early if we get to a nice stopping point :)

Anyone following along is obviously welcome... here's the hangout link: https://plus.google.com/hangouts/_/calendar/YmhzQGdtYWlsLmNvbQ.a8n4hhfd0pnccsgj6pmjp9o860?authuser=0

First off, I want to be really really clear about two meta-things:
1. This is a tough API to get right... esp the Inject/Extract piece where we want loose coupling, compile-time safety, AND a somewhat conventional, comprehensible API surface area. It's basically an overconstrained problem and we'll probably need to think hard about which of the above is most important since we're not likely to check all three boxes.
1. I'm grateful for this contribution and feel optimistic that we'll get something in (hopefully fairly soon) that works for Bloomberg's scale and codebase.

Regarding an agenda / sticking points:
- The biggest question in my mind is whether we want to have separate APIs for C++98 and C++11. My inclination is to say *yes*, as OpenTracing has always been primarily focused on the needs of the instrumenting engineer and secondarily focused on the needs of the Tracer implementor... but others may disagree and this is important enough to discuss live, I think.
- I recognize that people don't like `dynamic_cast`: who does, after all? But there are many ways to deal with that feeling, and the approach in this PR is merely one of them. I would like to discuss @jmacd's idea (referenced above, on a Go-related thread) about this, and also to examine the costs (mostly in terms of surprising patterns and cognitive overhead for OpenTracing users) of the template-driven solution in this PR. I have read the various things written on this PR and out-of-band over email and I remain unconvinced about the tradeoff here.
- There are some things coming down the pipe for OpenTracing more generally in Q1, and they have not been written down centrally (my fault). I will literally be doing that today and will link to my writeup from this PR, but TL;DR I would prefer to decouple (a) the core Tracer and Span interfaces, (b) the process of finding a global singleton Tracer, and (c) the process of finding an "active" Span via some pluggable mechanism (e.g., in thread-local-storage). This PR addresses (a) and (b) but does not address (c), and it would be good to at least talk through possible approaches that decouple all three.

Of course I am also happy to talk about anything else that people want to bring up, but those are the big things on my list. Once we've come to a consensus I am confident that we'll be able to get through the actual PR review process quickly, as all of the people on this thread are thankfully excellent engineers and the programming part is not going to be the sticking point :)

lookfwd:
Like a lot mentioning (c). I would also add - a bit -

(d) - what you do with the `Recorder` interface in Lightstep... I'm wondering if there's space for something that removes the overhead of "I have 10 tracers and they all do HTTP requests - but each one of them needs to have their own way to send the off-band HTTP requests (which in the case of Lyft is quite specific/async). Could Lyft say "for HTTP off-band requests, use this" through a standard API so that every tracer can use that optimal way

(e) Error handling. There's a class of errors that are related to an operation e.g. `getBaggage()` -> "baggage item not found" and things that are generic tracer errors (and could happen in any operation) e.g. "can't contact tracer". I have a feeling that an API should return only the 1st type of errors and have some callback - abstract mechanism to inform the application on tracer's generic errors. Tracing is an "aspect" in a sense. But it shouldn't bloat the app as much as possible. Having to wrap every `setBaggage()` or whatever operation like `handleStandardTracerErrors(setBaggage())` ... doesn't feel very good.

SaintDubious:
Thanks @bensigelman for setting up a conversation. I wanted to have a clear example of usage of this proposed interface so we could discuss. This is in response to the comment "surprising patterns and cognitive overhead for OpenTracing users." I'm confident that the **usage** of this interface is not surprising at all, in fact it looks almost identical to the usage of the previous API. Here's a sample:

```
#include <dtrace_tracer.h>
#include <text_map_carrier.h>

using namespace BloombergLP::dtrace;

int main( int argc, char** argv )
{
    // Create and store custom Tracer
    // To use another Tracer this is only change required
    TracerImplOptions opts;
    TracerImpl tracer( opts );
    TracerImpl::installImp( &tracer );

    // Make a top level span
    Span* span = Tracer::start( "top_level_span" );
    span->tag( "StringTag", "Hello" );
    span->tag( "IntTag", 22 );
    span->log( "logging", "Everything is awesome" );

    // Inject span into carrier
    TextMapCarrier carrier;
    Tracer::inject( &carrier, *span );

    // Extract span context and create child span
    // Usually this would be done in another service
    const SpanContext* span_context = Tracer::extract( carrier );
    SpanOptions* span_options = Tracer::makeSpanOptions();
    span_options->setOperation( "op" );
    span_options->setReference( opentracing::SpanReferenceType::e_ChildOf, *span_context );
    Span* child_span = Tracer::start( *span_options );
    Tracer::cleanup( span_options );    
    Tracer::cleanup( span_context );    
    child_span->finish();
    Tracer::cleanup( child_span );

    span->finish();
    Tracer::cleanup( span );

    return 0;
}
```

bensigelman:
No need to restart the grand debate (if you want to, you can have the last word!), but just so you know where I stand on it:

```
    Tracer::cleanup( span_options );
    Tracer::cleanup( span_context );
    ....
    Tracer::cleanup( span );
```

^^^ I consider these to be pretty problematic... def a gotcha.

I would also prefer that `::` be reserved for things that are logically static... IIUC, the syntax for using the global tracer ends up being different than the syntax for calling an impl directly. Which basically means that people have to use the global Tracer, which is not something that's been well-received as a *requirement* for OT use in other languages. I.e., it's fine to have a global tracer mechanism, but it shouldn't be the only way to use the sanctioned public API.

RomanDzhabarov:
@bhs can you send invite my way as well?

>> (d) - what you do with the Recorder interface in Lightstep... I'm
wondering if there's space for something that removes the overhead of "I
have 10 tracers and they all do HTTP requests - but each one of
>> them needs to have their own way to send the off-band HTTP requests
(which in the case of Lyft is quite specific/async). Could Lyft say "for
HTTP off-band requests, use this" through a standard API so that >> every
tracer can use that optimal way

I'd expect that there is a way to buffer/batch individual span creations
into a single request to collectors. As, for example, LS system is designed
to handle all of the spans/traces without any client sampling, sending
every single span in individual request seems to be problematic... cc: @bhs
With the batching it's going to be very little overhead having X tracers
sending HTTP requests.




--
Roman Dzhabarov
Software Engineer
[image: Lyft] <http://www.lyft.com/>

SaintDubious:
:) I'm so sorry, I know that by responding I'm totally grabbing the last word, but I'm actually really excited that both of @bensigelman's gotchas are minor (gives me hope that we are closer then I thought).

The cleanup functions are not required. @jquinn47 put those in to allow implementers the freedom to allocate in ways more interesting then just new/delete (for example a pool). The cleanup is very C++11 friendly as you can specify it as a delete function to a shared_ptr, which allows for the very clean usage (in fact our utility classes do exactly this with our in-house smart pointers):
```
// probably won't actually compile, but very close
std::shared_ptr<Span> sp( Tracer::start( "hi" ), Tracer::cleanup );
```

Also, the usage of :: is also not required, they were just added as helpers. A library user can use them to force creation of a Tracer even without access to main. But if the community prefers explicit creation I imagine that's not a major sticking point for us. Either way you can still use a Tracer pointer in the expected way with the proposed API, ie this works:
```
// We could create a globalTracer() like the existing API if people prefer
Tracer* tracer = Tracer::instance();
Span* span = tracer->start( "new_span" );
```
Okay, I'll go home now and stop posting. I'm very excited to speak to everyone on Tuesday.

bensigelman:
Vaguely related to my `(c)` above: https://github.com/opentracing/specification/issues/23#issuecomment-274223802

PS: @SaintDubious glad to read your comment above... helpful!

jquinn47:
I'd agree the static interface seems heavy handed. The `static` may be acceptable for the globally installed tracer, but it's overkill; @bensigelman is in the right here.

 (@SaintDubious, I know I have a problem, but it's interesting and I can't help it 👍)

lookfwd:
@RomanDzhabarov On the overhead, I don't mean computational and +1 for the batching of course! What I mostly mean is that I like the way LightStep's Recorder is used [here](https://github.com/lyft/envoy/blob/009d341ac8c6d1f70ceaa725b907086dc20587d9/source/common/tracing/http_tracer_impl.cc#L213). What it does, is that it expresses to the Tracer implementation, the right way, in the context of a given application, to send off-band data through http. This is very useful. Now let's assume that you have a Zipkin C++ Tracer and it implements OT. Given the lack of standardisation on the recorder, Zipkin, written with other developers, will likely by default use a (likely blocking) library (e.g. CURL) to send HTTP and will force you to link it. Then you will have to hack the Zipkin implementation to allow it to (also) allow a Recorder pattern that fits the Envoy async way of working. I think nevertheless, it's highly likely you would still have to link Zipkin's other HTTP sending mechanism. This whole mess is the overhead I'm talking about. I think from what I see [here](https://github.com/lyft/envoy/blob/009d341ac8c6d1f70ceaa725b907086dc20587d9/source/common/tracing/http_tracer_impl.cc#L176), an API could look roughly like this:

    struct HttpRecorder
    {
        struct Message
        {
            virtual ~Message();
            virtual const string & getUrl() const = 0;
            virtual const string & getMessage() const = 0;
            virtual const vector<string> & getHeaders() const = 0;
            virtual void onSuccess() = 0;
            virtual void onSuccess() = 0;
        };

        virtual ~HttpRecorder();
        virtual void sendHttpMessage(Message * message) = 0;
    };

jquinn47:
I do like @jmacd's idea. For C++ at least, I would take it one step farther and enrich the existing Span interface to require `externalize(map<string, string>&)` or `externalize(vector<char>&)`. 

By doing so, we can specialize the type system but always have a reliable fallback. None of this would require runtime checks on behalf of the implementor. C++ would do it for us:

```
class Span {
    // all of the current span methods ...
    virtual void externalize(map<string, string>& ) = 0;
};

struct RedSpan: public Span{
  void externalize(map<string, string>& m){ m["id"] = std::to_string<int>(_id); }
  int _id;
};

struct BluSpan: public Span{
   void externalize(map<string, string>&m) { m["bid"] = _id;
   string _id;
};

struct Carrier{
    virtual void accept(Span*) = 0;
};


struct CarrierImp{
    virtual void accept(Span * s){ 
      map<string, string> m; 
      s->externalize(m); 
  } 
   virtual void accept(RedSpan *rs) {
       string& id = rs->_id;
   }
};
```

This was one of my original thoughts when tackling the API, but it required that the Span interface be enhanced. I thought the spec was pretty rigid at the time, and wasn't sure how welcome an addition to the base span interface would be, but it does clean up some things nicely.

I'm not convinced I've thought through all of the ramifications yet. I'm most concerned about when we have different tracers interoperating, but I'll save all of that for Tuesday I suppose

jquinn47:
Upon further review, I'm not sure my previous idea would work either. Span's would still need about all the carriers and vice versa. The physical organization of the libraries would still be a problem :(

jquinn47:
Submitted a virtual inheritance rough draft in #7. Trade offs:

Pros: 
* Traditional OO
* Gains ability to multiplex

Cons:
* Carrier code is harder to get write (minor)
* More expensive 'no-ops' (minor)

I didn't want to close this PR, but I'd hate to split the discussions across to of them. I'll let the opentracing gurus decide where the discussions should continue.

bensigelman:
First, to summarize what was a fun and productive hangout today with @jquinn47, @SaintDubious, @lookfwd, @yurishkuro, @RomanDzhabarov, @jmacd, @tedsuo, and myself:
- Maintaining both C++98 and C++11 OpenTracing libraries is problematic, especially for (C++11) processes that mix libraries that depend on both.
- C++11 features can be built as a wrapper around a C++98 core.
- Concerns were raised about the template-driven approach... mainly cognitive overhead and Tracer multiplexing. Hence @jquinn47's speedy turnaround with #7.
- We spent a while talking about custom formats+carriers and how to manage dependencies sanely along those lines.

Anyway, @jquinn47: thank you for #7! IMO it's a worthwhile tradeoff. I feel odd cutting the (valuable) discussion in this PR off... let's wait for other opinions about #6 vs #7 and we can go from there.

SaintDubious:
Thank you @bensigelman for arranging the hangout yesterday, that was extremely useful. And thank you @jquinn47 for quickly giving us a new API to consider. 

My attempt to summarize the sticking point over Carriers and Spans:
- Ben considers the ability to select Span type at runtime to be a feature
- We consider this same exact ability to be a design flaw

For us, there is no way to change Span types over our system. There are too many services, transport mechanisms, and languages, there is no realistic way to upgrade them all. So we have to enforce exactly **one** Span type.

With the Virtual approach we have to play policy police with every Carrier implementer and say "You MUST cast this virtual Span to a concrete type and you may NEVER use anything else." Eventually some team will decide that within their island of services they are going to use some OTHER Span type and eventually that will leak out and the whole system becomes unstable.

We feel that the virtual approach allows for a flexibility that is dangerous to the system. It would be very helpful for us if someone has some real world examples of when they might want to use this feature.

lookfwd:
The way I perceived it, there was no definite "Span (or anything) at runtime has to be there." What I felt was mostly that "there's a "cognitive overhead" on templates.. the more we can narrow their usage, the better. Let's keep the `virtual` approach (main benefit: simplicity) still a bit open till we are definitely convinced that there's no way to specify correct and generic enough implementations by using it."

> There is no way to change Span types over our system

To clarify this - I guess that `Span`s will be evolving/changing over time but we can't imagine a single service switching `Span` types at run-time or that two types of `Span`s might co-exist in a system. Is that what you mean @SaintDubious?

lookfwd:
One more thing I would like to clarify for the entire discussion, that, I believe, is quite C++ specific. We have to set the expectations right for the OT C++ spec/library. Please spend a few seconds to have a look on an implementation of the C++ `std::string` type [here](http://www.sgi.com/tech/stl/string).

I guess my point is quite clear. That's the source code/header file of the `std::string` that everybody thinks that understands and absolutely everybody uses. If you ask people what's `std::string` - nobody will have that in mind. Highly likely they have the documentation page from [here](http://www.cplusplus.com/reference/string/string/) in mind or some chapter from Josuttis's book.

I think that implicitly, and partially because of other languages like Java, Go and Python, we have an unrealistic expectation on what *good, simple and quality* C++ code looks like.

Even large companies might have C++ experts that don't use templates, and that's perfectly fine and expected and lovely for *application code*. Not for library code. For example let's have a quick look on google test/google mock e.g. [here](https://github.com/google/googletest/blob/master/googlemock/include/gmock/gmock-matchers.h). I can't see anything that looks like what's [so nicely described in the documentation](https://github.com/google/googletest/blob/master/googlemock/docs/CheatSheet.md) and I seriously doubt anyone could create C++ library code that meets google mock's objectives by using just inheritance.

yurishkuro:
If I understand correctly what you mean by "changing Span type", the situation mostly arises with higher order tracers. These are some scenarios where higher order tracers are needed:
  1. the organization decides to switch from one tracing system to another. Given that all microservices cannot be upgraded simultaneously, there's a period when the org needs to run two tracing systems concurrently, so a multiplexing tracer can be used that dispatches the instrumentation calls to two underlying (actual) tracers
  1. tracing instrumentation is used for purposes beyond tracing, e.g. for emitting RPC metrics. The simplest way to do that is to create a decorator tracer that emits metrics and then delegates to the underlying real tracer
  1. somewhat similar to (1) - we want to run a demo application (like Ben's donut salon or Jaeger's HotROD) with more than one tracer, to demonstrate the OT's main proposition that instrumentation stays independent of the tracing backend.

On the topic of custom carriers, as always I find it hard to follow abstract discussions about what could happen and prefer to take concrete use cases and see how they are solved in the proposed API. The specific use case for custom carrier I have is the following. Uber uses tchannel, a kind-of RPC framework & binary wire protocol, which has built in support for zipkin-like tracing at the binary format level. Using those message slots is more efficient than passing tracing data via headers. So the instrumentation in various tchannel libraries (in different languages) perform inject/extract as a two-step process
  1. first they try to inject/extract using a custom carrier that represents tracing context as 4 fixed numeric fields. If the tracer understands that Format and carrier type (as Jaeger's tracer does), then the more efficient path is chosen by the instrumentation to encode those fields into the message.
  1. if the tracer says it does not support that format/carrier (in Go there's an unsupported carrier format error code that tracer can return), then the instrumentation uses TextMap format/carrier to encode the span context as arbitrary key/value pairs that are then stored in the message headers.

There was a sticky point about the interface of the carrier. In Go it wasn't much of a problem due to duck typing. In Python/Java we simply used a dict/Map as the carrier to pass the 4 fields.

Another concrete examples of custom carriers often arrises in Java, where various http frameworks use different types to represent http request. So what most instrumentations do is use the TextMap (or rather HttpHeaders) OT format with a wrapper around the specific framework's http request type that masquerades that request object as the standard text map OT carrier.

jquinn47:
Trying to keep the conversation here, but this is a response to @yurishkuro comment on #7:

This is rehashing what we discussed yesterday. 

The _problem_ with the approach your suggesting, in a statically built language like C++, is that we have a problem the way the libraries are physically laid out.  If I have a tracer that needs to know about different carriers, today, we're asked to write this:

```
#include <redcarrier.h>
#include <bluecarrier.h>

Tracer::inject(Carrier * c, span* s){
   if(RedCarrier * rc = dynamic_cast<RedCarrier*>(c)){
    ... 
   }
  else if (BluCarrier * bc = dynamic_cast<BlueCarrier*>(c){
    ... 
  } 
  else if (TextMapCarrier * tmc = dynamic_cast<TextMapCarrier*>(tmc){
     ...
  }
}
```

This library is going to be built one of two ways:

1: The tracer defines the carriers in its implementation
2: The tracer uses carriers defined outside of its implementation

In **both** situations, we have real world build problems with this library at scale.

In case 1, If I have an application that only sends 'Blue' messages, but I want the tracer implementation, I'm now forced to link against All of the different carrier libraries that the tracer needs. 

Using your concrete example, If you only had tchannel in use in your applications, but the tracer implementation also supported protobuf-over-tcp carriers, would you want to link against that library?


In case 2, if the tchannel carrier is defined outside of the tracer implementation, the tracer still needs to link against it. 

Now we have link link lines that start to look like: `-ltracer -ltchannel -lprotobuf-over-tcp -ltracer`.
This circular dependency is a rats nest, and clients are **still* forced to link against a whole bunch of libraries, they'll probably never need.

Now what I did in #7 was try to do what @jmacd  was suggesting. Push those details into the carriers, so that the link lines look more appropriate for your task `-ltracer -ltchannel`. Since `-lprotobuf-over-tcp` is never used in your app, you don't need to link against it.

jquinn47:
Now, ignoring the CRTP vs virtual debate, I believe their is a fundamental problem with the spec.

I have not yet seen how a migration path should work for switching out tracer implementations at scales. A lot of our discussion has been to about features used to accomodate this use case, but I don't see a clean way to pull it off. The demos are nifty to test a corner of the interface, but I'd posit they're not practical  at scale, once multiple applications begin communicating with each other over a set protocol. 

Say we're using `zipkin` today. We have installed it into 10k+ applications, all of which are built and deployed on different time frames. Now we want to switch to `jaeger`, for whatever reason.

For this to work, we first need a multiplexing tracer. All of the original tasks need to be rebuilt with a multiplexor first, but only using `zipkin`. We need to do this so the tracer implementation can start to understand what it means to receive multiple in-band representations of a span over the wire.

Now once everyone has that linked in, we can start to make `jaeger` optional, and continue to require `zipkin` information. The semantics of what to do if one tracer works but another fails are still undefined, but lets assume that they're both working. 

Once that's done, we then need to require `jaeger` is required, but we can start to mark `zipkin` details as optional.

Once `zipkin` is done, we can officially drop `zipkin` support and move back to a single tracer, only supporting `jaeger`.

This is a nightmare to pull off.  This is the line in the [spec](http://opentracing.io/documentation/pages/api/cross-process-tracing.html) that is resonating the most with me

> There is no expectation that different OpenTracing implementations Inject and Extract SpanContexts in compatible ways. Though OpenTracing is agnostic about the tracing implementation across an entire distributed system, for successful inter-process handoff it's essential that the processes on both sides of a propagation use the same tracing implementation.

Since using multiple tracers has significant maintenance side effects, and there is no reason to assume more than one tracer is operating system wide to begin with, I begin to ask, why is all of this flexibility put in place to begin with. I'm going to try my hardest to avoid the scenario I've described.

I hope this helps drive the discussion. I really don't care about CRTP vs virtual to be honest. The semantics of how these tasks are upgraded and interoperate are loosely defined. That's what I'm most interested in tightening up. I believe the design of the classes would come naturally once the problem is well formed and we're all talking about the same thing.

yurishkuro:
> I have not yet seen how a migration path should work for switching out tracer implementations at scales.

As Ben mentioned on the call, should this be considered the fault of the spec or the nature of the problem itself? How would you solve this if there was no OpenTracing, and you had 10k apps instrumented with bespoke `zipkin` API and you wanted to switch to another tracing system with completely different libraries and a completely different API? Now I think *that* would be intractable problem to solve. If all your 10k apps are instrumented with OpenTracing, then the process you described is painful, but it's doable.

Fundamentally there are only two ways to solve this problem as far as I can see, either (a) follow the process you described and go through a migration period when two tracers are running in the apps, or (b) try to make the two tracing implementations interoperate. In this specific example (ironically), Jaeger tracers have the ability to read incoming Zipkin's B3 headers, and Jaeger backend has the ability to accept Zipkin data model for spans, so one could get away with interop instead of dual tracers. But that's only because Jaeger was originally built on parts of Zipkin and we had similar data models and wire propagation format. If you're trying to swap tracing systems that have more significant differences (very simple example: one uses uint64 trace IDs, the other string IDs), then interop is not possible, and dual-deployment migration is the only path.

Also, if we avoid extreme examples of 10k apps (per 80-20 rule the majority of companies have much fewer microservices, say less than a 100), then the dual-tracer approach becomes a lot more feasible and not as time consuming.

jquinn47:
I'd agree with your last point. I'm am (for better or worse) apart of the 20% there

yurishkuro:
> Trying to keep the conversation here, but this is a response to @yurishkuro comment on #7:

You mean this comment: https://github.com/opentracing/opentracing-cpp/pull/7#issuecomment-275166046. I do not believe it introduces any library linkage issues, because the custom carrier is defined purely in the specific RPC library and the tracer does not need to know about it. The carrier in that case is just an adapter to transform the internal representation of the framework's remote request (like servlet.HttpRequest) to the interface required for the Format, like TextMapReader/TextMapWriter in Go. I am not sure if CRTP can do that, I'd think you need virtual methods for that. This usage of custom carriers is by far the most prevalent. Even Go OT uses it as part of the official API, see [HTTPHeaderCarrier](https://github.com/opentracing/opentracing-go/blob/master/propagation.go#L158), which wraps http.Header as plain TextMap carrier.

The other example I gave, with TChannel, is indeed where the library linkage issue comes into view. That is not just a custom carrier, but also a custom Format. This is very rare use cases as far as I know, and we managed to solve it without having cross-dependencies by agreeing to use standard (widely available) interface for the carrier itself, e.g. in Java we'd simply use `Map<String, String>`. The custom part here is really just the format, which says "this data has special semantics and you can use it more efficiently in your specific protocol than if this was a TextMap format". This requires tight coupling of the RPC framework with the capabilities (but not interfaces) of the tracer. And the reason it's so rare is because it implies that the transport has special abilities to carry tracing information that is different from standard key/value metadata. In most cases these special abilities already imply some dependency on the tracing data representation. But as far as OT API is concerned, it's nice to have this capability available.

lookfwd:
To make this discussion more hands-on. I implemented an ugly throw-away draft for:

* a C++11 wrapper that I believe that might bring satisfactory answer to the question if we will be able to provide something smooth for C++11 people
* a Higher Order Tracer that supports two tracers and by recursion any number (yeah :D) which raises lots of interesting issues but I also thing that will bring the discussion back to more hands-on stuff. The main file:

```
#include <app_tracing.h>

int main()
{
    // Setup
    ZipkinV1 t1;
    ZipkinV2 t2;

    MyHotTracer imp(t1, t2);

    GlobalTracer::install(&imp);


    {
        // Application code
        using Tc = Cpp11GlobalTracer;

        Tc tracer = Tc::instance();

        std::shared_ptr<Tc::Span> span = tracer.start("hi");

        std::vector<opentracing::TextMapPair> pairs;

        tracer.inject(&pairs, *span);

        std::shared_ptr<const Tc::SpanContext> ctx = tracer.extract(pairs);
    }

    // Cleanup
    GlobalTracer::uninstall();

    return 0;
}
```

To compile & run:

    g++ -I. -D__USE_XOPEN2K8 -std=c++11 main.cpp hot.cpp app_tracing.cpp build/opentracing/libopentracing.a -o main && ./main

lookfwd:
To make this discussion more hands-on. I implemented [here](https://github.com/jquinn47/opentracing-cpp/pull/1/files) an ugly throw-away draft for:

* a C++11 wrapper that I believe that might bring satisfactory answer to the question if we will be able to provide something smooth for C++11 people
* a Higher Order Tracer that supports two tracers and by recursion any number (yeah :D) which raises lots of interesting issues but I also thing that will bring the discussion back to more hands-on stuff. The main file:

```
#include <app_tracing.h>

int main()
{
    // Setup
    ZipkinV1 t1;
    ZipkinV2 t2;

    MyHotTracer imp(t1, t2);

    GlobalTracer::install(&imp);


    {
        // Application code
        auto tracer = Cpp11GlobalTracer::instance();

        auto span = tracer.start("hi");

        std::vector<opentracing::TextMapPair> pairs;

        tracer.inject(&pairs, *span);

        auto ctx = tracer.extract(pairs);
    }

    // Cleanup
    GlobalTracer::uninstall();

    return 0;
}
```

To compile & run:

    g++ -I. -D__USE_XOPEN2K8 -std=c++11 main.cpp hot.cpp app_tracing.cpp build/opentracing/libopentracing.a -o main && ./main

jquinn47:
I don't want to speak too much for @SaintDubious  and @lookfwd, so if they disagree, prepare to hear back from them :)

I've personally gone full circle on this in my head at least 3 times now. Re-reading @yurishkuro comments, I may be looking at what I see as problems too harshly through my 'bloomberg' goggles. I'm still seeing issues with regards to where in the code Span implementations are exposed, and which components are responsible for knowing about those details. I've got many teams to convince to get this spread out, and if we can't come to a consensus here, I believe it will be even more difficult as I add more cooks to the opentracing kitchen.

We (myself, @SaintDubious , and @lookfwd ) will have to be picky about the first tracer we deploy at large. We will be experimenting on a small scale over the next few weeks. (Likely with the CRTP interface since we already have it, then replacing it) to iron out a few kinks with other teams and see what speed bumps are waiting for us. CRTP has already shown itself to be problematic; we hit a template compiler bug on solaris and have seen otherwise experienced developers complain about the cognitive overload that @bensigelman was wary of. While @lookfwd , myself, and probably STL implementors may be fine with it, we're likely in the minority. In my opinion, those are enough red flags to nix it.

I would love to see what @jmacd does in the `go` version. I think I'm going to hold off on moving forward with the virtual version until I get a better grasp of the approach he is taking there.

I'm certain the implementation we move forward with is going to need to be "version aware" and we only make changes in our organization in a backwards compatible way. That's my problem to deal with and not necessarily OT's, but if the spec is still very fluid that would make adoption harder. 

I was hoping the spec would have a recommendation for that, but @yurishkuro arguments are convincing. That's likely an intractable problem to ask a spec to solve.

lookfwd:
@jquinn47, after our last chat, I came to appreciate another dimension in complexity i.e. service ownership. That was the "final nail" and I'm perfectly convinced that there's a layer of OT that we will have to adopt and stick with it, unchanged for a couple of years. It's impossible to coordinate all those people to even relink and redeploy. I don't think we should miss all the developments and the opportunities in the tracing industry because of this though... "All problems in computer science can be solved by another level of indirection" and this brings the concept of OT "virtualisation" I vaguely mentioned (continued after the image).

enterprise-ot

You have a very lightweight Tracer that "talks OT v.1.0" (and will likely for a decade) i.e. does no transformation or anything clever whatsoever. Maybe it could have some tiny or moderate (configurable) intelligence around sampling but that's about it. All those data get forwarded to an OT proxy that must be installed on every server. You can update and configure it however you like. It instantiates as many OT Tracers as it needs to talk to any implementations and those Tracers might be of any version of OT. It might have versioning logic that decides routing or where something should be forwarded and where not. You own this so you do whatever you want. Would something like this solve the problem?

bensigelman:
@adriancole: the propagation interop pain point is painful for *you* as someone who maintains a tracing system that is dealing with version skew across numerous forks/etc. The lack of standard tracing instrumentation is the pain point I hear over and over again when speaking with (50+? 100+?) engineering orgs that are integrating tracing systems.

TL;DR it seems you are conflating the needs of people who spend their time implementing tracing systems with the needs of people who spend their time trying to integrated with them.

OpenTracing does indeed separate the concern of in-band propagation formats from the concern of describing the semantics of application software. In fact, that's mostly the point. As software complexity grows, we need some common way of describing behavior while making minimal assumptions about consumers of those descriptions. But indeed: distinct tracing systems have distinct in-band propagation formats (even distinct *versions* of tracing systems have distinct in-band propagation formats).

FWIW, I agree wholeheartedly with @adriancole about the red-flag-ish-ness of talking about "10 years" with any API that is in its early stages. I feel relatively comfortable making "conceptual" guarantees about most of the core abstractions in OT, but true compile-time compatibility between whatever is in this PR (or #7) and `opentracing-cpp` circa 2027 is a very different story. Of course I am motivated to think through this API, and taking future migrations into account is a great way to do that, but let's be clear about the fact that it will almost certainly change.

adriancole:
>
> @adriancole <https://github.com/adriancole>: the propagation interop pain
> point is painful for *you* as someone who maintains a tracing system that
> is dealing with version skew across numerous forks/etc. The lack of
> standard tracing instrumentation is the pain point I hear over and over
> again when speaking with (50+? 100+?) engineering orgs that are integrating
> tracing systems.
>
> TL;DR it seems you are conflating the needs of people who spend their time
> implementing tracing systems with the needs of people who spend their time
> trying to integrated with them.
>
Version skew and numerous forks (to the degree that's even true) is no more
a part of my focus today than any time I've been outside a monorepo. I
*happen* to also do a workshop occasionally, for implementors *because* I
spend most of my time with integrators. Integrators show up organically in
my world. I don't go company to company.. employees of companies (and
random people) show up from every time zone. It isn't organized.

My goal was to highlight that propagation is a worthwhile thing to focus on
as if it is busted, your traces are busted. In other words, I'm encouraging
people to accept vs perpetually reject negative feedback about decisions
made.

bensigelman:
For what it's worth, I don't go company to company either, Adrian... I probably should but have been too busy. Most of these conversations I've had are ad hoc things.

It's a little hard to read the tea leaves, but I think (?) you are suggesting that I'm the "people" who should "accept" rather than "perpetually reject" negative feedback. If so, I always welcome feedback, positive or negative, about anything I'm involved with.

That said, if that "feedback" is actually in the form of *advice* to 3rd parties, it seems rather important to present the other side of the argument. (Which, in this case, is about whether (a) instrumentation of application code or (b) consistency about in-band propagation formats is the sticking point for distributed tracing deployments in the wild)

adriancole:
>
> It's a little hard to read the tea leaves, but I think (?) you are
> suggesting that I'm the "people" who should "accept" rather than
> "perpetually reject" negative feedback. If so, I always welcome feedback,
> positive or negative, about anything I'm involved with.
>
The challenge I'm making is indeed pointed at you, not because of your
opinion topic is being discussed, rather you are the person who offers
meetings in response to people questioning things.

My 2p is that when something comes up where people are participating in
something like C++, but not privy to the historical reasons on something,
go ahead and find or open an issue for them! This issue description could
include the rationale which is otherwise lost being held across N issues.
This also helps others who aren't watching the threads like this or able to
attend F2F meetings.

There's already a spec repo, so most of this is just applying it
https://github.com/opentracing/specification/issues

The only really new part is the explicit idea that progress is possible
(i.e. read api past 1.0.. like a github tag or milestone), so that fresh
people have a role in shaping the future.

lookfwd:
"10 years" <- This will happen anyway... it would happen with any API we would create internally or Open Source

> "conceptual" guarantees

That's the main reason I was pushing towards using OT instead of something proprietary. OT is way more mature conceptually compared to any v1.0 we would make in-house in such short time.

This proxy ideas above, are such that we can decouple instrumentation code on apps that might be stuck in v1.0 for long time and be able to evolve with the OT API (after the proxy) without having to relink/deploy all apps.

SaintDubious:
After much internal discussion, @jquinn47 and I have decided to not press for this PR anymore. So unless someone really wants to argue for it (which I strongly doubt) it is safe to close this. While we continue to like the static polymorphism and concrete type of the span when passed to the carrier, the CRTP is annoying to work with (as @bensigelman feared it would be).

bensigelman:
@SaintDubious @jquinn47 @lookfwd 

Thanks for the note, I was wondering how your conversation was proceeding. So what do you think about something like #7?

lookfwd:
@bensigelman I like the direction of https://github.com/opentracing/opentracing-cpp/pull/7 with the proxy note above but we should have more solid evidence in ~2-weeks time.

tedsuo:
Closing this issue, as the discussion has moved on to #7.

PR #12, opened 2017-08-27
Fix CMakeLists.txt for case-sensitive file systems.

rnburn:
CMakeLists.txt uses the wrong capitalization when including CPack.


lookfwd:
Looks good :)

PR #14, opened 2017-08-29
Extremely basic cmake find module file to allow the use of find_package.

pantoss:
I'm not entirely sure where to put this cmake find module but I thought others might find it useful.

### DIFF
--- a/c++11/CMakeLists.txt
+++ b/c++11/CMakeLists.txt
@@ -37,16 +37,42 @@ install(DIRECTORY include/opentracing DESTINATION include
 
 set(SRCS src/propagation.cpp src/noop.cpp src/tracer.cpp)
 add_library(opentracing SHARED ${SRCS})
+target_include_directories(opentracing INTERFACE "$<INSTALL_INTERFACE:include/>")
 set_target_properties(opentracing PROPERTIES VERSION ${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}
                                              SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR})
 add_library(opentracing-static STATIC ${SRCS})
 set_target_properties(opentracing-static PROPERTIES OUTPUT_NAME opentracing)
+target_include_directories(opentracing-static INTERFACE "$<INSTALL_INTERFACE:include/>")
 if (CLANG_TIDY_EXE)
   set_target_properties(opentracing PROPERTIES
                                     CXX_CLANG_TIDY "${DO_CLANG_TIDY}")
 endif()
-install(TARGETS opentracing opentracing-static
+
+
+install(TARGETS opentracing opentracing-static EXPORT OpenTracingTargets
         LIBRARY DESTINATION lib
         ARCHIVE DESTINATION lib)
+
+# Package configuration setup
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file("${CMAKE_CURRENT_BINARY_DIR}/OpenTracingConfigVersion.cmake"
+    VERSION ${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}
+    COMPATIBILITY AnyNewerVersion)
+export(EXPORT OpenTracingTargets
+    FILE "${CMAKE_CURRENT_BINARY_DIR}/OpenTracingTargets.cmake"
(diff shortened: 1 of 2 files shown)


lookfwd:
How does one use this? Ideally where should it be placed?

rnburn:
I think this would be helpful, but could we have it more closely follow approach outline [here](https://cmake.org/Wiki/CMake/Tutorials/How_to_create_a_ProjectConfig.cmake_file).

pantoss:
lookfwd: Without installing it as part of CMake you would have to do something like.
set(opentracing_dir /path/to/where/this/cmake/config/file.cmake)
find_package(opentracing)

Then you can use opentracing_include_dir and opentracing_library in your project. You can also use open trace_root_dir to each for the headers and libraries in non-standard locations I believe.

rnburn: I think this describes a more up to date approach. https://cmake.org/cmake/help/git-master/manual/cmake-packages.7.html#creating-packages  Ideally I would like have it do version checking. Unfortunately I find cmake maddening so I went for the simplest approach which seemed to work.

rnburn:
@pantoss If we're going to add something like this in, I think it should support version checking and also install the config file to a standard location -- that seems like it's the convention for project configs.

pantoss:
@rnburn So I've tried to follow the documents on how to do this properly and amazingly it seems to work. This is my first foray into GitHub so i'm not sure what I should do, should I close this and do another pull request?

rnburn:
@pantoss Thanks. Just leave this PR open. We can always squash the commits when merging. I'll review it more closely later.

pantoss:
So I've had another go. I've tried to resolve all your points. 
- Hope I didn't miss any places with the lower case T
- I've changed the namespace to camel case but left the target as lower case to match the library. I wasn't really sure which way around was better.
- Sorted formatting I think. At least it looks better than it did. Ashamed I hadn't done this already.
- Hope the comments are clear enough and cover what you require.

It seems that the modern method doesn't set the <package>_INCLUDE_DIR and <package>_LIBRARIES variables. All you get is the namespace targets. I've asked about and the advice was, if it is new package then there isn't any point in adding them because there is nothing to be backwards compatible with. If you require them I think I can put them in.

Let me know what you think.

c++11/CMakeLists.txt
```diff
@@ -37,16 +37,42 @@ install(DIRECTORY include/opentracing DESTINATION include
...
+write_basic_package_version_file("${CMAKE_CURRENT_BINARY_DIR}/OpenTracingConfigVersion.cmake"
+    VERSION ${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}
+    COMPATIBILITY AnyNewerVersion)
+export(EXPORT OpenTracingTargets
+    FILE "${CMAKE_CURRENT_BINARY_DIR}/OpenTracingTargets.cmake"
+    NAMESPACE OpenTracing::)
+configure_file(cmake/OpenTracingConfig.cmake
+    "${CMAKE_CURRENT_BINARY_DIR}/OpenTracingConfig.cmake"
+    COPYONLY)
+set(ConfigPackageLocation lib/cmake/OpenTracingConfig)
```
rnburn: Can we install to `lib/cmake/OpenTracing` instead? That's the convention that the example uses.

pantoss:
I hate cmake as much as I like it, by next week this approach will be out dated also. Thanks for being so thorough. I'll keep it in mind but I'm going to stick with the current form for the time being.

pantoss:
I've read a bit more and I believe the approach you suggest is intended for creating a module which is distributed separate from the source. But I might be wrong. It doesn't seem very nice compared to the current method where you simply export the target and the final location of the installation is correctly determined. You can treat the target libraries just like  any other library target and you can also have multiple targets. I haven't played with it and might be miss understanding it. it does explicitly say that is should be used over the configure_file but I think it is meaning when creating a package for an external source.

pantoss:
I missed your comment about the location of the installation of cmake files. I've corrected this now.

c++11/cmake/OpenTracingConfig.cmake
```diff
@@ -0,0 +1,12 @@
+# - Locate the opentracing headers and libraries
+# Usage:
+#    find_package(OpenTracing)
+#    find_package(OpenTracing 0.1.0)  # To specify a minimum version
+# This will create the targets:
+#    OpenTracing::opentracing # dynamic library
+#    OpenTracing::opentracing_static # static library
```
rnburn: The documentation name `OpenTracing::opentracing_static` doesn't match the target name `OpenTracing::opentracing-static`.

Issue #15, opened 2017-09-06
Cannot compile as std++14 in Xcode


pantoss:
When using this in a project which is c++14 code I get the following error:

In file included from /usr/local/include/opentracing/propagation.h:5:
In file included from /usr/local/include/opentracing/util.h:13:
/usr/local/include/opentracing/expected/expected.hpp:829:19: error: no matching member function for call to 'construct_error'
        contained.construct_error( error.value() );
        ~~~~~~~~~~^~~~~~~~~~~~~~~
/usr/local/include/opentracing/expected/expected.hpp:1206:16: note: in instantiation of function template specialization 'opentracing::expected<void, std::__1::error_code>::expected<std::exception_ptr>' requested here
        return make_unexpected_from_current_exception();
               ^
/usr/local/include/opentracing/expected/expected.hpp:157:10: note: candidate function not viable: no known conversion from 'const std::exception_ptr' to 'const error_type' (aka 'const std::__1::error_code') for 1st argument
    void construct_error( error_type const & e )
         ^
/usr/local/include/opentracing/expected/expected.hpp:162:10: note: candidate function not viable: no known conversion from 'const std::exception_ptr' to 'error_type' (aka 'std::__1::error_code') for 1st argument
    void construct_error( error_type && e )
         ^
1 error generated.

Apple LLVM version 8.1.0 (clang-802.0.42)
Target: x86_64-apple-darwin16.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

rnburn:
Could you attach the smallest file you can that reproduces the problem?

pantoss:
This is the simplest I can make it.

#include <opentracing/tracer.h>

int main()
{
    return 0;
}

Compiled with:
clang++ -isystem /usr/local/include  -g   -std=gnu++14 -o main.cpp.o -c main.cpp

pantoss:
This doesn't look well formed to me:

auto make_expected_from_call( F f,
    nsel_REQUIRES( ! std::is_same<typename std::result_of<F()>::type, void>::value )
) -> expected< typename std::result_of<F()>::type >
{
    try
    {
        return make_expected( f() );
    }
    catch (...)
    {
        return make_unexpected_from_current_exception();
    }
}

make_expected must return a expected<T, std::error_code>
make_unexpected_from_current_exception has to return expected<T, std::exception_ptr>

[...]

PR #16, opened 2017-09-06
Fixes for expected.

rnburn:
* Adds in a missing header (See https://github.com/martinmoene/expected-lite/issues/12).
* Removes functions that break when the default error type for expected is changed from `std::exception_ptr` to `std::error_code` (#15).


tedsuo:
This looks fine... @jmacd any reason you see not to merge?

PR #18, opened 2017-09-20
Remove C++98 directory.

rnburn:
This PR removes the C++98 subdirectory so as to make the library easier to version. If both C++98 and C++11 libraries were included in the same repository it would make tagging for releases difficult to manage.

I also changed the required cmake version from 3.0 to 3.1 because that's [needed](https://stackoverflow.com/a/31010221) by CMAKE_CXX_STANDARD.


rnburn:
@heyleke Thanks. I added a note about the cmake version bump.

PR #19, opened 2017-09-25
Store tags in StartSpanOptions as string. Correct comment.

pantoss:
This change should reduce the chance of dangling pointers in a number of scenarios. It also allows for moving the tags from the StartSpanOptions directly into the span, which is probably a common use case when constructing a span. I did do a small test and this was faster because of allowing move semantics. It would also be possible to use move semantics in the construction of StartSpanOption implementations which might further further improve performance.

Corrects a comment which gives an example of use.

### DIFF
--- a/include/opentracing/tracer.h
+++ b/include/opentracing/tracer.h
@@ -63,12 +63,12 @@ class Tracer {
   //     // The vanilla child span case:
   //     auto span = tracer.StartSpan(
   //         "GetFeed",
-  //         {opentracing::ChildOf(parentSpan.context())})
+  //         {opentracing::ChildOf(&parentSpan.context())})
   //
   //     // All the bells and whistles:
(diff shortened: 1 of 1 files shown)


rnburn:
Spans wouldn't be able to move tags from StartSpanOptions since StartSpanWithOptions takes it by [const ref](https://github.com/pantoss/opentracing-cpp/blob/184ea684637c3eeea56aed1d34ff6b86184cb367/c%2B%2B11/include/opentracing/tracer.h#L89), but I think that's ok. Allowing move semantics would require adding an overload for `StartSpanOptions&&` which I'm not sure is worth the complexity.

rnburn:
Thanks @pantoss. Can you just sync this with the changes that went in for the directory structure. It looks good otherwise.

PR #21, opened 2017-10-01
Allow user to disable testing.

isaachier:


rnburn:
Thanks for the PR. The changes look good.

Could you just s/ENABLE_TESTING/BUILD_TESTING/? That seems like it's the more standard [way](https://cmake.org/cmake/help/v3.1/module/CTest.html) of controlling whether tests are build.

isaachier:
Oh sorry my mistake. Meant the latter.

PR #25, opened 2017-10-16
Support key lookup in span context extraction.

rnburn:
Some concerns have come up around the efficiency of extracting a span context. The current API forces tracer implementations to iterate and compare against all the key values in a carrier. If there’s no indexing on the keys, this is reasonable; but when there’s key indexing, it requires the tracer to do more comparisons than are necessary.

This PR adds optional support for looking up values by key. A carrier can either support lookup for all, some, or no keys; and it doesn’t break compatibility with the existing API. Tracers are still required to support extracting span context using the [ForeachKey method](https://github.com/opentracing/opentracing-cpp/blob/master/include/opentracing/propagation.h#L103), but can also take advantage of the lookup API for faster extraction if available.

### DIFF
--- a/include/opentracing/propagation.h
+++ b/include/opentracing/propagation.h
@@ -80,6 +80,15 @@ const std::error_code invalid_carrier_error(2, propagation_error_category());
 const std::error_code span_context_corrupted_error(
     3, propagation_error_category());
 
+// `key_not_found_error` occurs when TextMapReader::LookupKey fails to find
+// an entry for the provided key.
+const std::error_code key_not_found_error(4, propagation_error_category());
+
+// `lookup_key_not_supported_error` occurs when TextMapReader::LookupKey is
+// not supported for the provided key.
+const std::error_code lookup_key_not_supported_error(
+    5, propagation_error_category());
+
 // TextMapWriter is the Inject() carrier for the TextMap builtin format. With
 // it, the caller can encode a SpanContext for propagation as entries in a map
 // of unicode strings.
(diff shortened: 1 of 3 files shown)


include/opentracing/propagation.h
```diff
@@ -89,6 +98,18 @@ class TextMapReader {
...
 
+  // LookupKey returns the value for the specified `key` if available. If no
+  // such key is present, it returns `key_not_found_error`.
+  //
+  // TextMapReaders are not required to implement this method. If not supported,
+  // the function returns `lookup_key_not_supported_error`.
+  //
+  // Tracers may use this as an alternative to `ForeachKey` as a faster way to
+  // extract span context.
+  virtual expected<string_view> LookupKey(string_view /*key*/) const {
```
yurishkuro: I'm super rusty in C++, but doesn't requiring the tracer implementation to run in try-catch introduces its own overhead? Exceptions should be used for exceptional situations, it's not an error or exception for the carrier to not support by-key lookup. I'd rather have something like `optional<KeyedTextMapReader> getKeyedReader()`

rnburn: This is an error code approach: it doesn't use exceptions.

In some of the projects that might use this (e.g. envoy, nginx), they index some of the keys but not all of them; so I tried to design it in a way that key-value lookup is supported on a per-key basis, not just a simple on/off for all keys.

pantoss: Wouldn't it be better to have this as an additional interface, which implementations can inherit? e.g. 
    class Indexable
    {
     public:
         virtual optional<string_view> operator[](string_view) const = 0;
    };

This way if you want to test if it is supported you can do a dynamic_cast. If you are some how certain that it will be supported you can just do a static_cast(I think) to avoid any RTTI overhead. You don't have to handle the not supported case in every call to LookupKey either, which will have its own overhead, even if it is just instruction cache use or branch prediction.

rnburn: @pantoss Lookup needs to be supported to be on a per-key level. Some applications don't index all of the header keys. For example, envoy provides fast lookup only for a list of [hard-coded keys](
https://github.com/envoyproxy/envoy/blob/master/source/common/http/header_map_impl.h#L102).

Issue #26, opened 2017-10-20
C API


isaachier:
Not sure if this has ever been addressed, but would it make sense to include a C API header? I was going to try and write one for my own client, but realized it probably makes more sense to wrap the OpenTracing API than my own custom API. Just to reference an example of what I mean, see LevelDB's C header in [c.h](https://github.com/google/leveldb/blob/master/include/leveldb/c.h). Adding a header like that for the OpenTracing API would simplify writing bindings for C based languages such as Python and Lua.

rnburn:
I agree it would be useful. The issue of an OpenTracing C API has come up a few times on [gitter.im/opentracing-cpp](https://gitter.im/opentracing/opentracing-cpp) and there were some people interested (@heyleke).

I think the plan was to define an OT C API in a separate repository and include some sort of bridge from the C++11. I don't think anyone was against such an API -- it's just that nobody has done the work of proposing one yet.

isaachier:
OK great. I may follow up on this with a proposal when I have the time.

isaachier:
It's been a while, but I think I'm almost ready to submit a proposal for a C OpenTracing API. Any ideas where to submit this? I don't think opentracing-c exists yet. Here is the link for now: [opentracing-c](https://github.com/isaachier/opentracing-c).

yurishkuro:
If people think it will be easier to discuss as a PR, I can create an official repo.

One immediate question I have:

```c
typedef struct opentracing_tracer {
    opentracing_span* (*start_span)(struct opentracing_tracer* tracer,
                                    const char* operation_name);
```

This looks like you're pretending to have an "object with methods" - is this a usual practice in C? Any other projects use this technique?

isaachier:
Ya to be fair these aren't truly methods because I'm passing the instance as the first argument each time. The C compiler just sees these as struct members that happen to be function pointers. Go structs do this too from time to time, but C has to use it exclusively. I think I first saw this in the CPython implementation but it's in [libevent](https://github.com/libevent/libevent/blob/master/include/event2/http_struct.h#L122)as well.

What I definitely saw first in CPython was the notion of implementing and interface by making it the first member of the derived struct (see https://github.com/python/cpython/blob/master/Include/object.h#L101). Since opentracing is heavy on interfaces, both techniques are used quite a bit.

mikz:
It is quite a pity there is no C API. 

[...]

PR #27, opened 2017-10-24
C API wrapper proposal

isaachier:
Motivation covered in #26. Similar to LevelDB C API wrapper.


rnburn:
Thanks @isaachier. You might also want to post it to the [opentracing-cpp](https://gitter.im/opentracing/opentracing-cpp) gitter channel. 

One question I have: If someone wanted to use tracers that were completely written in C as opposed to using this as a bride to the C++11 API, does this still give them a way to customize the tracer used?

isaachier:
@rnburn that is a much more difficult proposal. It would probably need another project `opentracing-c`. If anything, the C++ API should implement/extend the C API if that were the goal.

rbtcollins:
What I'd really like to see for C and C-derived OT APIs is a specification for dlopening tracers, so that we can get e.g. nginx (C) to buy-in just once to OT, and then not have to rebuild it to use different tracer implementations. Perhaps the C++ bindings already permit that.

rnburn:
@rbtcollins There's another issue (#28) to discuss just that.

The current instrumentation for [nginx](https://github.com/opentracing-contrib/nginx-opentracing) doesn't require you to rebuild nginx for each tracer as it uses nginx's dynamic modules, but there is some configuration overhead that we're trying to remove with `dlopen` conventions for tracers.

isaachier:
I would say there is at least a hint to this issue in the C API I proposed here. When it comes to initializing a global tracer, there is no way to specify this properly in C. That is why I left it as `TODO`. The `dlopen` solution would potentially resolve this.

isaachier:
For those looking for a true C implementation of the OpenTracing model, I would say gRPC's Census library comes pretty close to the specification (if not completely fulfills it). Take a look here: https://github.com/grpc/grpc/tree/master/src/core/ext/census.

src/c.cpp
```diff
@@ -0,0 +1,358 @@
...
+        return 0;
+    } catch (...) {
+        return -1;
+    }
+}
+
+extern "C" void opentracing_close_tracer(opentracing_tracer_t* tracer)
+{
+    tracer->impl_->Close();
+    delete tracer->impl_;
```
rnburn: `tracer->impl_` points to values managed by `std::shared_ptr`'s, so won't this result in freeing memory twice?

src/c.cpp
```diff
@@ -0,0 +1,358 @@
...
+            return result.error().value();
+        }
+        sc->impl_ = result->release();
+        return 0;
+    } catch (...) {
+        return -1;
+    }
+}
+
+extern "C" void opentracing_close_tracer(opentracing_tracer_t* tracer)
```
rnburn: Shouldn't there also be an `opentracing_free_tracer` function? `Tracer::Close` may wait for spans to flush, so it's not always desirable for applications to call it when freeing resources since it could make the program take much longer to terminate.

src/c.cpp
```diff
@@ -0,0 +1,358 @@
...
+    for (auto i = 0; i < options->num_references; ++i) {
+        const auto* ref = options->references[i];
+        cppOptions.references.emplace_back(
+                static_cast<opentracing::SpanReferenceType>(ref->type),
+                ref->context->impl_);
+    }
+
+    for (auto i = 0; i < options->num_tags; ++i) {
+        const auto* tag = options->tags[i];
+        cppOptions.tags.emplace_back(tag->key, tag->value->impl_);
```
rnburn: Functions like this can throw `std::bad_alloc` which wouldn't be caught anywhere.

src/c.cpp
isaachier: I noticed this issue this morning. This part of the implementation needs to be revisited.

src/c.cpp
isaachier: Interesting distinction. I can add free separately.

src/c.cpp
isaachier: You are right. I'll wrap everything in try blocks to avoid that.

isaachier:
The difficulty with wrapping the API stems from a few issues relating to ownership. One difficulty is that I cannot copy any of the classes because there is no `clone` method for me to use. I'm starting to wonder what a pure C API would look like. It seems the main consumers of pure C would much rather have the implementation be pure C as well.

rnburn:
I think you could fix most of the ownership issues by just changing `opentracing_tracer_t` to
```cpp
struct opentracing_tracer_t {
     std::shared_ptr<opentracing::Tracer> impl_;
 };
```

rnburn:
Hey @heyleke, @jcchavezs -- you guys have asked about a C API before. Would something like this work for you?

isaachier:
Maybe... I'll get back to this soon enough. Thank you for reviewing it.

jcchavezs:
I am in!

rnburn:
@jcchavezs Could you describe your use case? I'm curious what types of projects would benefit from this API.

isaachier:
@jcchavezs I am also curious to hear. Meanwhile, what motivated me was to ease the process of wrapping/binding other languages to a C/C++ implementation. That way, an implementation need only follow the opentracing-cpp specification, and benefit from a version in Python, Ruby, JavaScript, etc. with no additional code (or minimal code for each).

heyleke:
@rnburn  I'm following this with great interest. Although this approach will probably not solve the biggest issue in our use case: C++11 support in toolchain is still required. But having a start towards a C api to start with is already something, I also read about the census implementation which could be useful to create a pure C implementation if we can't solve the C++11 support.

So, the main use case are:
* pure C applications, but linking to C++ is still possible in that case
* environments lacking C++ or C++11 support.
(although I have to admit in 2017 you should be capable of doing C++11 in your toolchains ;-) )

Already alot of thx to @isaachier for bringing this on the table. I hope to become more active in this area in the near future.

Btw, if there one day would be a pure c implementation, I would call it opentracing-C99 if it would be pure C99 (e.g.)
And the C 'wrapper' would reside in the opentracing-cpp(11) repo.

Are there any pure C tracers that this would be testable with (with some modifications)? lightstep?

isaachier:
I just submitted a new version in pure C. As of now uses C11, although that
is just for the single atomic pointer used to store the global tracer.

On Oct 27, 2017 5:48 AM, "Jan Heylen" <notifications@github.com> wrote:

@rnburn <https://github.com/rnburn> I'm following this with great interest.
Although this approach will probably not solve the biggest issue in our use
case: C++11 support in toolchain is still required. But having a start
towards a C api to start with is already something, I also read about the
census implementation which could be useful to create a pure C
implementation if we can't solve the C++11 support.

So, the main use case are:

   - pure C applications, but linking to C++ is still possible in that case
   - environments lacking C++ or C++11 support.
   (although I have to admit in 2017 you should be capable of doing C++11
   in your toolchains ;-) )

Already alot of thx to @isaachier <https://github.com/isaachier> for
bringing this on the table. I hope to become more active in this area in
the near future.

Btw, if there one day would be a pure c implementation, I would call it
opentracing-C99 if it would be pure C99 (e.g.)
And the C 'wrapper' would reside in the opentracing-cpp(11) repo.

Are there any pure C tracers that this would be testable with (with some
modifications)? lightstep?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/opentracing/opentracing-cpp/pull/27#issuecomment-339926446>,
or mute the thread

rnburn:
@heyleke - Could you use a C API wrapper if the C++11 tracer and opentracing libraries were cross compiled and distributed as binaries?

isaachier:
If I understand correctly, the answer would be no. Certain platforms just don't have good support for C++11 (in my experience AIX and Solaris).

rnburn:
The native compilers for those OSs might have somewhat flaky C++11 support, but GCC/Clang support cross compiling to most of the widely used architectures/environments.

I believe he has an embedded target. I'm wondering if packages build with something like [ellcc](http://ellcc.org/) might solve the problem.

isaachier:
Any idea what to do with this issue or should I just close it?

rnburn:
I'd like to have more input from stakeholders and some example project instrumentations before going forward with something for C.

I expect there are some projects that would need a C API, but I don't think there's that many. The major C projects that I've looked at so far (NGINX, Varnish, HAProxy) can all be instrumented with the C++ API by just providing some `extern "C"` hooks.

isaachier:
Ya that's my feeling. The C problem is interesting but might be a separate project.

ringerc:
Raising my hand here for the PostgreSQL project. I don't know yet if it's practical to do it with our own `extern "C"` hooks, but it sure is a big barrier to adoption having to work around calling into C++.

isaachier:
I'm still not sure what the issue is here. If we had `extern "C"` hooks in OpenTracing and some way of initializing the global tracer without using C++ code, then the user will never need to deal with C++ code at all. The only valid complaint I have heard regarding this approach is that the resulting binary will be slightly larger because it will need to link in parts of the C++ runtime. Other than that, I see no reason to rewrite everything in C when there is no concrete need for it.

ringerc:
Exactly. I personally don't care about linking C++. Bigger binary? Meh, whatever. Symbol table bloat? Use visibility annotations when needed, and otherwise, meh, again not a big concern.

The C++11 toolchain is a harder sell, but so long as tracing can be conditionally built or you can link to a stub implementation instead or whatever, then who cares? No c++11, no tracing. Fine.

What I'm personally struggling with is what is needed to use this from C. It's so purely C++ I don't even know where to start. Presumably I have to implement some C++11/postgres glue and a few `extern "C"` entry points to call from the main postgres code. The docs are sparse though.

I'm  starting to see that this proposal was for a complete C binding for opentracing, implemented on top of the C++11 one. And instead the advice seems to be to write your opentracing integration in C++ and expose a narrow app-specific API as `extern "C"` to pass in the info needed to trace various points of interest.

Seems viable if potentially unpleasant in performance terms.

isaachier:
OK glad to hear that you aren't totally against it. If there's enough interest, I'm sure OpenTracing isn't against a C implementation. The issues you mention may be better solved in a particular tracing implementation, instead of the interface exposed in OpenTracing. For example, I worked on https://github.com/jaegertracing/cpp-client where we expose a number of samplers to allow for customization of the sampling procedure. OpenTracing is purposely minimal to allow for many potential implementations.

rnburn:
I reopened #26 as there've also been some people interested in using C bindings to reuse C++ tracers to instrument other languages.

I'm sure it's something we'll do at some point, and we'd welcome a PR for it if it's something you want to work on.

Combining a C wrapper with the dynamic loading proposal (#45), would give you a way to handle the C++11 dependency: Set up the symbols from the OT library to have late binding and load the tracer library in dynamically. If no tracing library is ever loaded, then the symbols won't be called and can remain undefined. And if a tracer library is loaded in, they'll be resolved by `dlopen`.

rnburn:
Also, there was some discussion initially on whether to have the C API be a wrapper to the C++ library or a completely separate OT C API that would allow you to build a tracer completely in C (which is what @isaachier latest iteration on the PR was).

I think it would be best to just focus on a C wrapper, as a complete OT C API would add additional complexity and I'm not sure I see anyone writing a tracer library in C anytime soon.

isaachier:
Agree we should avoid pure C as long as possible.

For some work I started for binding opentracing-cpp in python, see here: https://github.com/isaachier/opentracing-cpp-python. One main issue I faced was inheriting from opentracing-python (still not sure how possible this is).

rbtcollins:
I think having the C API be pure C expecting to call C routines - and then a C to C++ thunk driver - would provide more flexability - e.g. if someone wanted to write a rust driver, or to bind to a golang driver, they would be able to do so at the C ABI, not C++.

isaachier:
The problem with C is that almost everything would be from scratch. For example, what would be an ideal string class? `char *`, `const char*`, `struct strbuf`? Even if you figure those out, how do you implement inheritance? How do you implement classes? The idea of "upcasting" to a base class pointer sounds like a massive pain compared to using C++ classes out of the box.

If someone wants a C++ API without the standard library, I would be much more willing to hear it out. That would be very C-like and suffer from the same lack of common constructs (i.e. strings, vectors), but would at least have some standard way of implementing a class hierarchy.

PR #30, opened 2017-11-02
Remove default timestamping.

rnburn:
Also document StartSpanOptions' values.


rnburn:
Yes, no tests were affected.

PR #36, opened 2017-11-08
Add examples for Inject/Extract.

rnburn:


rnburn:
Merging in since this only changes documentation.

PR #41, opened 2017-11-21
Ensure that assert is not ignored for tests.

rnburn:
Right now tests aren't run if cmake is built in release mode because asserts are ignored. This PR makes sure `NDEBUG` is undefined in the testing code.


rnburn:
replaced with custom assert

rnburn:
@isaachier -- are you ok with this now?

PR #42, opened 2017-11-21
Support propagation directly to/from strings.

rnburn:
Adds functions to support propagation directly to/from strings so as to allow tracers to use more efficient implementations.

Includes default implementations for the binary string Inject/Extract functions that use `iostream`'s so that tracers aren't required to provide implementations for them.

Addresses efficiency [concerns](https://github.com/envoyproxy/envoy/pull/2017#discussion_r150085606) that came up when instrumenting envoy for OpenTracing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,7 @@ endif()
 include_directories(include)
 include_directories(SYSTEM 3rd_party/include)
 
-set(SRCS src/propagation.cpp src/noop.cpp src/tracer.cpp)
+set(SRCS src/propagation.cpp src/noop.cpp src/in_memory_stream.cpp src/tracer.cpp)
 add_library(opentracing SHARED ${SRCS})
 target_include_directories(opentracing INTERFACE "$<INSTALL_INTERFACE:include/>")
 set_target_properties(opentracing PROPERTIES VERSION ${OPENTRACING_VERSION_STRING}
--- a/include/opentracing/tracer.h
+++ b/include/opentracing/tracer.h
@@ -106,10 +106,13 @@ class Tracer {
   //
   // Throws only if `writer` does.
   //
-  // If `writer` is an `std::ostream`, then Inject() propagates `sc` as a blob
-  // of binary data.
+  // If `writer` is an `std::ostream` or std::string, then Inject() propagates
+  // `sc` as a blob of binary data.
   virtual expected<void> Inject(const SpanContext& sc,
                                 std::ostream& writer) const = 0;
+
+  virtual expected<void> Inject(const SpanContext& sc,
+                                std::string& writer) const;
 
   virtual expected<void> Inject(const SpanContext& sc,
                                 const TextMapWriter& writer) const = 0;
@@ -132,6 +135,9 @@ class Tracer {
   // Throws only if `reader` does.
   virtual expected<std::unique_ptr<SpanContext>> Extract(
       std::istream& reader) const = 0;
+
+  virtual expected<std::unique_ptr<SpanContext>> Extract(
+      opentracing::string_view reader) const;
 
   virtual expected<std::unique_ptr<SpanContext>> Extract(
       const TextMapReader& reader) const = 0;
--- a/src/in_memory_stream.cpp
+++ b/src/in_memory_stream.cpp
@@ -1,0 +1,15 @@
+#include "in_memory_stream.h"
+
+namespace opentracing {
+BEGIN_OPENTRACING_ABI_NAMESPACE
+InMemoryBuffer::InMemoryBuffer(const char* data, size_t size) {
+  // Data isn't modified so this is safe.
+  auto non_const_data = const_cast<char*>(data);
+  setg(non_const_data, non_const_data, non_const_data + size);
+}
+
+InMemoryStream::InMemoryStream(const char* data, size_t size)
+    : InMemoryBuffer{data, size},
+      std::istream{static_cast<std::streambuf*>(this)} {}
+END_OPENTRACING_ABI_NAMESPACE
+}  // namespace opentracing
--- a/src/in_memory_stream.h
+++ b/src/in_memory_stream.h
@@ -1,0 +1,21 @@
+#pragma once
+
+#include <opentracing/version.h>
+#include <istream>
+#include <streambuf>
+
+namespace opentracing {
+BEGIN_OPENTRACING_ABI_NAMESPACE
+// Class to allow construction of an istream from constant memory without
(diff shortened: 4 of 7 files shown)


yurishkuro:
> so as to allow tracers to use more efficient implementations

Not sure I understand. (a) Binary format is not supported by many implementations. (b) What is the meaning of using a binary codec to write to an `ostringstream` - won't you end up with garbage?

rnburn:
(a) Isn't binary one of the required propagation formats for OpenTracing? It's supported in the other [API's](https://github.com/opentracing/opentracing-go/blob/master/propagation.go#L52) as well. Envy has a custom single-header propagation mode that uses a base64 encoding of the binary span context representation.

(b) `stringstream`'s can be used with arbitrary binary data. They don't have to correspond to ASCII or any other printable format. (See https://stackoverflow.com/q/47125387/4447365)

yurishkuro:
> (a) Isn't binary one of the required propagation formats for OpenTracing? It's supported in the other API's as well.

Yes, it is "required" yet often not implemented because there are hardly any transport protocols where binary is actually required. Pretty much all instrumentation for HTTP protocol uses HTTP_HEADERS format, sometimes TEXT_MAP format is used for protocols that guarantee that keys and values do not get mangled.

> Envy has a custom single-header propagation mode that uses a base64 encoding of the binary span context representation.

Hm, that makes it incompatible with anything that's not going through Envoy. E.g. if you have a remote client (mobile or maybe an API client) talking to the backend over HTTP, the normal way for that client to pass the context is using the tracer's HTTP_HEADERS format.

I'm not suggesting you stop doing this, it's just an interesting development I wasn't aware of. The OT way is pretty much "tracer rules them all", there is no standard on how the context is represented on the wire, not even which headers are used. The [TraceContext Spec](https://github.com/TraceContext/tracecontext-spec/) is almost complete opposite, it standardizes both the headers _and_ the values. And Envoy's approach is in the middle, standard header, but 100% custom value.

yurishkuro:
question directly related to the PR though - why extend the tracer API with `string&` writer instead of putting that code in Envoy?

rnburn:
I added an option in the OT PR so that users can configure their tracer to propagate using it's native format if they want, but the custom format is still there since Envoy has been using that for a while.

Their propagation method was discussed quite a bit [here](https://github.com/envoyproxy/envoy/issues/1711) for context. With Zipkin, they actually do propagation twice, both in their custom format and Zipkin's native B3 format.

rnburn:
> question directly related to the PR though - why extend the tracer API with string& writer instead of putting that code in Envoy?

The point string overloads is so that tracer's might implement those functions directly to propagate more efficiently. The code I provided is just a fallback to iostream Inject/Extract if tracer implementations don't want to bother with also writing Inject/Extract for strings.

yurishkuro:
> The point of the string overloads is so that tracer's might implement those functions directly to propagate more efficiently. 

But you are transposing Envoy's non-standard (from OT perspective) approach into OT API. There is no standard format in OT for "string blob". If the other side of the connection is an OT tracer from the same tracing system but in different language, it won't be able to deal with the string blob.

yurishkuro:
thanks for the link, btw - I saw the beginning of that discussion, but not how it ended (should've subscribed).

rnburn:
Yes, that is a drawback of how they're propagating. I think at Lyft they do this form of custom propagation in their other languages as well; so provided the binary formats in the different languages match, they can still propagate.

Nothing about their single header format is leaking into this PR, though. It's only about allowing for the already supported binary format to be more efficient.

rnburn:
For reference: [this](https://github.com/rnburn/envoy/blob/master/source/common/tracing/opentracing_driver_impl.cc#L91) is how the proposed Envoy OT PR propagates. It supports both Envoy's custom format and the tracers native http format.

yurishkuro:
I'm just suggesting that you can do the exact same thing directly in Envoy - use the binary/ostream API but pass an ostringstream. No OT API in other languages define codecs for a string blob.  Efficiency here == utility ==> does not need to be part of the API (because it's confusing and unexplainable based on the OT spec).

I am not even sure why you need the string in the first place - base64 encoding is fundamentally defined over `[]byte` not strings (could be a quirk of the specific encode you're using).

rnburn:
Right, but `std::string` is the closest thing you have to a `[]byte` array in C++11. A separate `std::byte` type was only added in [C++17](http://en.cppreference.com/w/cpp/types/byte). Using `std::string` like this to package binary data is fairly standard [practice](https://stackoverflow.com/a/4655232/4447365).

Envoy's concern with `std::ostringstream` was that it [requires](https://stackoverflow.com/a/47115697/4447365) you to make unnecessary copies of the data. With `std::string`, a tracer can potentially propagate more efficiently by calculating how much space it needs and allocating memory only once.

rnburn:
For comparison, Google's C++ Protobuf library [provides functions](https://developers.google.com/protocol-buffers/docs/cpptutorial#parsing-and-serialization) that serialize directly to/from `std::string` as well as iostreams.

tedsuo:
This looks to me like an efficient implementation for binary propagation in C++. Can we focus on whether this is the correct way to implement binary in C++, and leave aside discussion of the utility of the binary transport?

yurishkuro:
It's not about the utility of the binary format, merely about the API design. We already have an `f(X)` in the API, but we're trying to add `g(Y)` which is trivially expressible in terms of `f`. So we're increasing the API surface without any new functional requirement, merely as a convenience, a convenience that can just as easily be implemented as a static utility function elsewhere.

rnburn:
@yurishkuro That's not quite true. The Inject/Extract functions are virtual, so tracer libraries can provide their own more efficient implementations if they want.

Since they can be expressed in terms of the iostream Inject/Extract functrions, the OT library does provide a default implementation so that libraries aren't required to implement them.

But the whole point of the PR is to allow for more efficient implementations if you're using a strings for binary propagation.

yurishkuro:
> so tracer libraries can provide their own more efficient implementations if they want.

Rule of Three - do we have 3 library providers who are interested in overriding this method in order to use more efficient implementation?

isaachier:
I tend to agree. I have seen too many C++ programmers worried about streams vs. strings and strings vs. string views/refs when that is the least of their performance concerns. My fear is that the C++ implementation will become too C-like if performance is given priority over elegance/simplicity of use. If there is a significant performance issue and/or it is a deal-breaker for those considering OpenTracing, I would reconsider.

rnburn:
I'll close this since envoy was ok merging in a version using iostreams.

But I do think that unnecessary memory allocations are absolutely something to be concerned about in the API.

Issue #35, opened 2017-11-08
Add log function for runtime types


isaachier:
As of now, a user must use an initializer list to log key-value pairs. The problem is, this effectively disables any dynamic logging. For example, if I want to log a list of key-value pairs, I have to know this list at compile time, seeing as no STL container can be converted into an initializer list at runtime.

Here are a few ideas for a new log function overload:

```
using Iterator = std::vector<std::pair<string_view, Value>>::const_iterator;
virtual void Log(Iterator first, Iterator last) noexcept = 0;
```

or

```
```

isaachier:
```
virtual void Log(const std::unordered_map<string_view, Value>& pairs) noexcept = 0;
```

isaachier:
```
virtual void Log(const std::unordered_map<string_view, Value>& pairs) noexcept = 0;
```

isaachier:
```
virtual void Log(const std::unordered_map<string_view, Value>& pairs) noexcept = 0;
```

isaachier:
```
virtual void Log(const std::unordered_map<string_view, Value>& pairs) noexcept = 0;
```

isaachier:
```
virtual void Log(const std::unordered_map<string_view, Value>& pairs) noexcept = 0;
```

isaachier:
```
virtual void Log(const std::unordered_map<string_view, Value>& pairs) noexcept = 0;
```

rnburn:
What would you think of adding an overload for something like [LogRecord](https://github.com/opentracing/opentracing-go/blob/master/span.go#L123) that's in the Go API?

 You can use it to customize the timestamp of the log as well.

isaachier:
I'd like that too. TBH, anything other than `std::initializer_list` is fine with me.

PR #43, opened 2017-12-07
Add LogRecord API from Go client

isaachier:
Fix for #35

### DIFF
--- a/include/opentracing/span.h
+++ b/include/opentracing/span.h
@@ -45,7 +45,7 @@ struct FinishSpanOptions {
   //
   // None of the LogRecord.timestamp values may be SystemTime() (i.e., they must
   // be set explicitly). Also, they must be >= the Span's start system timestamp
-  // and <= the finish_steady_timestamp converted to steady timestamp
+  // and <= the finish_steady_timestamp converted to system timestamp
   // (or SystemTime::now() if finish_steady_timestamp is default-constructed).
   // Otherwise the behavior of FinishWithOptions() is undefined.
   std::vector<LogRecord> log_records;
--- a/include/opentracing/span.h
+++ b/include/opentracing/span.h
@@ -8,6 +8,7 @@
 #include <chrono>
 #include <functional>
 #include <string>
+#include <vector>
 
 namespace opentracing {
 BEGIN_OPENTRACING_ABI_NAMESPACE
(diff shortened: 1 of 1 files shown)


rnburn:
@isaachier - Do you think it would make sense to also be able to log these directly on the span? (I know Go only supports it as part of FinishSpanOptions.)

include/opentracing/span.h
```diff
@@ -27,10 +28,27 @@ class SpanContext {
...
+
+  // log_records allows the caller to specify the contents of many Log() calls
+  // with a single vector. May be empty.
+  //
+  // None of the LogRecord.timestamp values may be SystemTime() (i.e., they must
+  // be set explicitly). Also, they must be >= the Span's start system timestamp
+  // and <= the finish_steady_timestamp converted to steady timestamp
+  // (or SystemTime::now() if finish_steady_timestamp is default-constructed).
+  // Otherwise the behavior of FinishWithOptions() is undefined.
+  std::vector<LogRecord> log_records;
```
rnburn: Could we change this to unspecified behavior? 

Also,
```
// and <= the finish_steady_timestamp converted to steady timestamp
```
I think you mean `converted to system timestamp` here.

isaachier: Correct, my mistake.

isaachier:
I was thinking the same thing but the Go version may do that logic intentionally. For example, if I log a record `A`, then add `B` and `C` in the `FinishSpanOptions`, does the implementation append those, and if so, are they appended to the end or sorted based on timestamp? For that reason, it may be worthwhile to stick to one approach unless those questions are resolved.

yurishkuro:
my advice is to keep api simple, don't introduce features nobody asked for.

rnburn:
@isaachier -- could you fix the comments for this so that we can merge this in?

isaachier:
Sorry yes will do now.

include/opentracing/span.h
```diff
@@ -27,10 +28,27 @@ class SpanContext {
...
   SteadyTime finish_steady_timestamp;
+
+  // log_records allows the caller to specify the contents of many Log() calls
+  // with a single vector. May be empty.
+  //
+  // None of the LogRecord.timestamp values may be SystemTime() (i.e., they must
+  // be set explicitly). Also, they must be >= the Span's start system timestamp
+  // and <= the finish_steady_timestamp converted to system timestamp
+  // (or SystemTime::now() if finish_steady_timestamp is default-constructed).
+  // Otherwise the behavior of FinishWithOptions() is undefined.
```
rnburn: > ```
> // Otherwise the behavior of FinishWithOptions() is undefined.
> ```
Can we change this to unspecified or remove that line. [Undefined behavior](https://www.quora.com/What-is-the-difference-between-undefined-unspecified-and-implementation-defined-behavior) would mean a tracer is free to do anything (abort, segfault, etc). Given that converting between steady and system time is a noisy process, I  don't think it would be right for a tracer to abort or corrupt the process if a timestamp is out of bounds.

PR #46, opened 2017-12-23
CMake: make shared/static target a configurable option

heyleke:
in order to make it available as a package in buildroot (https://buildroot.org).

See http://lists.buildroot.org/pipermail/buildroot/2017-December/209932.html

Br,

Jan Heylen


rnburn:
LGTM. Thanks!

Issue #50, opened 2018-01-25
compiler restriction for mocktracer are too strict


heyleke:
https://github.com/opentracing/opentracing-cpp/blob/359bafe4f5e026400620827c23e8e571ad875c02/mocktracer/3rd_party/nlohmann/include/opentracing/mocktracer/nlohmann/json.hpp#L67

see https://github.com/nlohmann/json#supported-compilers

I believe a library like opentracing-cpp cannot be as strict in supported compilers as mentioned here, IMHO the line should be c++11, which is AFAIU gcc 4.8.5 (with certain cpu architectures excluded).

So, either this check is wrong, or this component is not a good dependency.

rnburn:
Maybe switch to [rapidjson](https://github.com/Tencent/rapidjson)? It only requires C++98, but it's a much bigger library.

heyleke:
or maybe this one: https://github.com/taocpp/json although I'm no c++11 expert.

This might be a good reference:
https://github.com/miloyip/nativejson-benchmark

heyleke:
looks like taocpp/json is still in beta and not as fast as rapid.

rnburn:
I don't care to much about the performance since the mocktracer is only mean to support unit tests, but I'll get something figured out before I make the next release. (Should probably also add gcc 4.8.5 to the CI coverage).

heyleke:
correction: actually c++11 feature complete is 4.8.1, not 4.8.5 (that is just the one that ships with centos7)

rnburn:
I put in #54 that changes to do only serialization and uses manual code instead of a library.

I removed the deserialization code since that can always be done by the code testing against the mocktracer.

PR #51, opened 2018-01-26
travis: add 4.8.1 test

heyleke:
(not sure how I can test this without creating a pull request, so here it goes)


heyleke:
fail is expected, see https://github.com/opentracing/opentracing-cpp/issues/50

rnburn:
ok. thanks.

PR #53, opened 2018-02-01
Improve documentation.

rnburn:
* Fixes minor grammar issues.
* Documents how SpanContext `nullptr`s are handled.

### DIFF
--- a/include/opentracing/tracer.h
+++ b/include/opentracing/tracer.h
@@ -205,7 +205,9 @@ class StartTimestamp : public StartSpanOption {
 // referenced SpanContext. See the SpanReferenceType documentation for
 // supported relationships.
 //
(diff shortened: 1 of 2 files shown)


include/opentracing/tracer.h
```diff
@@ -202,6 +204,8 @@ class StartTimestamp : public StartSpanOption {
 // SpanReference is a StartSpanOption that pairs a SpanReferenceType and a
 // referenced SpanContext. See the SpanReferenceType documentation for
 // supported relationships.
+//
+// If the referenced SpanContext is a nullptr, it is ignored.
```
ringerc: Seems fine, also something like:

```
+ The passed SpanContext is copied during Span construction and the pointer is not retained.
```

rnburn: Added.

PR #54, opened 2018-02-11
Remove dependency on nlohmann json library.

### LINKED ISSUE #50: compiler restriction for mocktracer are too strict

rnburn:
Fixes #50


rnburn:
@heyleke -- look ok?

PR #56, opened 2018-02-20
Support copying span contexts.

rnburn:
Adds support for copying span contexts.

Follow up from https://gitter.im/opentracing/opentracing-cpp?at=5a82810018f388e62697ab54


rnburn:
@isaachier - Not a high priority, but could you take a look at this when you get the chance?

mocktracer/src/mock_span_context.cpp
```diff
@@ -21,12 +21,20 @@ void MockSpanContext::ForeachBaggageItem(
   }
 }
 
-void MockSpanContext::SetData(SpanContextData& data) {
+void MockSpanContext::SetData(SpanContextData& data) const {
```
isaachier: This name is a bit confusing. Sounds like you are setting the data member named `data`. Maybe `CopyData`?

rnburn: Sure. I'll rename it.

d4l3k:
@rnburn Do you know when this will make it into a tagged release? It'd be really nice to be able to use this

PR #88, opened 2018-05-28
Improve error handling for mocktracer.

rnburn:


mocktracer/test/tracer_factory_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+    span_filename.append(std::to_string(random_id));
+    std::string configuration = R"({ "output_file": ")";
+    configuration.append(span_filename);
+    configuration.append(R"(" })");
+
+    auto tracer_maybe =
+        tracer_factory.MakeTracer(configuration.c_str(), error_message);
+    REQUIRE(tracer_maybe);
+
+    REQUIRE(std::remove(span_filename.c_str()) == 0);
```
isaachier: Although you probably don't have anyone using AIX (C++11 support isn't great), I recall that file removal on that platform doesn't return `0` unless the creation was made with an exclusive ownership flag or something like that. Bottom line, you may not need to be so strict about file removal.

PR #89, opened 2018-05-28
Fix endian issue with mocktracer propagation.

rnburn:


rnburn:
yeah, I was going for something as simple and portable as possible.

PR #91, opened 2018-06-03
Abi versioning

### LINKED ISSUE #81: Conventions for ABI changes?
When an app loads a C++ tracer dynamically, all the C++ tracers currently do a comparison of the version of the OT API they were built with to the version of the OT API the app was built with. If there's any difference, they fail with a [incompatible_library_versions_error](https://github.com/opentracing/opentracing-cpp/blob/master/include/opentracing/dynamic_load.h#L70). This is meant to protect against using a tracer and app with incompatible ABIs. Any differences in the ABI can cause the app to segfault.

While this was probably ok behavior to start with, I think it may be too strict as it will cause loading to fail for even minor non-ABI-breaking changes.

I'm wondering if we can establish some conventions for ABI changes to allow for less strict checking.

Perhaps require any ABI change to update at least the minor version of the repository so the comparison can ignore the patch version?

And for reference, these are examples of some changes and how they would affect the ABI:
* Adding, removing, or reordering data members for structs like StartSpanOptions, SpanReference, etc. (Breaks ABI)
* Adding, removing, or reordering  virtual functions. (Breaks ABI).
* Adding a non-virtual function. (Doesn't change ABI).

rnburn:
Sets up conventions for versioning the ABI to allow for less strict checking when dynamically loading tracers.

* Changes the `OpenTracingMakeTracerFactory` function prototype to take an ABI version in addition to an API version. `OpenTracingMakeTracerFactory` should fail only if the ABI version is different. (The API version is still passed in case there's ever some sort of change that breaks semantic compatibility).
* Establishes conventions for incrementing the ABI version and marking unstable when between releases.

I also plan to put in an additional PR with some CI coverage that will help to detect ABI breakage.

Fixes #81


CMakeLists.txt
```diff
@@ -5,7 +5,14 @@ project(opentracing-cpp)
 # ==============================================================================
 # Version information
 
-set(OPENTRACING_ABI_VERSION "1")
+# Increment ABI version for any ABI-breaking change.
+#
+# Also, whenever the ABI is between versions and in development
+# suffix the ABI version number with "_unstable".
```
isaachier: Might be worth adding this comment to some sort of release documentation. Jaeger repos usually have a `RELEASE.md` for this type of thing.

test/multiple_tracer_link_test.cpp
```diff
@@ -0,0 +1,3 @@
+// Contains nothing, but links in tracer_a.o and tracer_b.o to verify that
+// there's no multiple definition error from OpenTracingMakeTracerFactory.
```
isaachier: I wonder if this is guaranteed to cause a compilation error. Might be worth using the function pointer for some trivial operation to ensure the symbol isn't discarded during linking.

mocktracer/src/dynamic_load.cpp
```diff
@@ -4,29 +4,34 @@
...
-  *tracer_factory =
-      new (std::nothrow) opentracing::mocktracer::MockTracerFactory{};
-  if (*tracer_factory == nullptr) {
-    *error_category = static_cast<const void*>(&std::generic_category());
-    return static_cast<int>(std::errc::not_enough_memory);
-  }
+  *tracer_factory = new opentracing::mocktracer::MockTracerFactory{};
 
   return 0;
+} catch (const std::bad_alloc&) {
```
isaachier: Out of curiosity, why'd you change to `try`/`catch`? If you expect other types of exception, make sure to catch `std::exception` or `catch (...)`.

rnburn: I added the try/catch because it assigns to the string `error_message`, which can allocate memory. `std::bad_alloc` should be the only exception that could be thrown.

CMakeLists.txt
rnburn: Sounds good. I'll add a `RELEASE.md`.

rnburn: Added.

test/multiple_tracer_link_test.cpp
rnburn: Done.

Issue #85, opened 2018-05-22
Set up CI testing for windows.


rnburn:
AppVeyor was set up for the repository (See https://github.com/opentracing/opentracing-cpp/issues/80), but the build tests still need to be written.

mdouaihy:
I am working on it. I should have something in few hours.

cwe1ss:
Just let me know if you need anything from the AppVeyor account!

hypnoce:
@mdouaihy you found the issue for crests ?

mdouaihy:
it's taking a bit more time than expected because MSVC handle the shared libraries differently than Linux. Also, MSVC requires to declare the exported and imported symbols using declspec(export) and declspec(import) while compiling the shared lib but not when compiling the static lib, etc. I should have a working version running and tested on Appveyor by the end of this week end

rnburn:
@mdouaihy - how's this going?

It looks like your branch also adds dynamic loading support for windows -- that's cool; but could we maybe break that out into a separate PR and put in a smaller PR that just adds windows CI support without dynamic loading to get started?

mdouaihy:
@rnburn, I took time before submitting any PR because in parallel, I worked on building Jaeger on Windows also (https://github.com/mdouaihy/jaeger-client-cpp).

As for your suggestion, here are the problems:
* Unit tests were not building
** The first change i made fixed two compile errors build does not fix the link of the unit tests
** With Visual Studio, dynamic libraries should be moved next to the unit test binaries
* Unit tests are failing. 
** unordered_map does not have the same order between MSVC and GCC/Clang
** there is a unit test for dynamic loading. If I dont add the dynamic support, the test will fail;

hypnoce:
Great work from @mdouaihy who worked hard on it !

rnburn:
@mdouaihy - thanks for the update.

This [check](https://github.com/opentracing/opentracing-cpp/blob/master/test/CMakeLists.txt#L20) should disable the dynamic loading test if it's not supported. Is it not working?

Which test does the unordered_map cause problems for? I can try to fix it.

rnburn:
@mdouaihy - I put in https://github.com/opentracing/opentracing-cpp/pull/92 to fix the JSON ordering issue.

PR #92, opened 2018-06-11
Make mocktracer JSON consistent

rnburn:
This PR replaces `unordered_map` with `map` so that the JSON representation for the mocktracer 's spans will be consistent.

See https://github.com/opentracing/opentracing-cpp/issues/85#issuecomment-396086818


rnburn:
@mdouaihy - Thanks. The mocktracer is really only meant for testing, so it doesn't matter too much how efficient it is.

isaachier:
In my [C client](https://github.com/isaachier/jaeger-client-c) I originally used an array for baggage, but implemented a hashtable in the end because baggage can be used as a generic transport mechanism so it can grow quite large.

PR #93, opened 2018-06-11
Support build on Windows with MSVC 15.7 (#85)

mdouaihy:
* Export Symbols
* Add Runtime targets to the install steps. On windows, static libraries are very tricky. shared library is needed.
* Add some MSVC compile options (_SCL_SECURE_NO_WARNINGS). This warning warns about unsafe methods or usage of the stl.
* Add appveyor build file. Credits for @hypnoce
* Add output directory and ignore build folder in git


rnburn:
Thanks @mdouaihy. I'll get to reviewing this soon.

include/opentracing/dynamic_load.h
```diff
@@ -37,10 +38,10 @@
...
 extern "C" {
 #ifdef OPENTRACING_BUILD_DYNAMIC_LOADING
-int __attribute((weak))
-OpenTracingMakeTracerFactory(const char* opentracing_version,
-                             const void** error_category,
-                             void** tracer_factory);
+OPEN_TRACING_EXPORT int OPEN_TRACING_ATT
+    OpenTracingMakeTracerFactory(
+    const char* opentracing_version, const void** error_category,
+    void** tracer_factory);
```
rnburn: `OPEN_TRACING_ATT` doesn't expand to anything on windows, right? 

Suppose I have two tracing libraries libTracerA, libTracerB, and then I link them both into my app with the intention of switching between them based on a config -- would that give a multiple definition error on windows because both would define `OpenTracingMakeTracerFactory `?

The weak attribute was there to support that scenario. Could we maybe use [`__declspec(selectany)`](https://msdn.microsoft.com/en-us/library/5tkz6s71.aspx) for this?

include/opentracing/symbols.h
```diff
@@ -0,0 +1,39 @@
...
+#include <opentracing/config.h>
+
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4251)
+#endif
+
+#ifdef _MSC_VER
+
+#define OPEN_TRACING_EXPORT __declspec(dllexport)
```
rnburn: Could we s/OPEN_TRACING/OPENTRACING/ to keep the naming convention consistent?

src/dynamic_load_windows.cpp
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  std::unique_ptr<DynamicLibraryHandle> dynamic_library_handle{
+      new DynamicLibraryHandleWindows{handle}};
+
+  const auto make_tracer_factory =
+      reinterpret_cast<decltype(OpenTracingMakeTracerFactory)*>(
+          GetProcAddress(handle, "OpenTracingMakeTracerFactory"));
+  if (make_tracer_factory == nullptr) {
+    error_message = "An error occurred whike looking up for OpenTracingMakeTracerFactory : " + GetLastError();
```
rnburn: > An error occurred whike

typo

include/opentracing/dynamic_load.h
mdouaihy: Indeed, OPEN_TRACING_ATT does not expand to anything on Windows.

if we take your example, why would I link directly with the two tracers? I'd rather use the dynamic loading mechanism to load the implementation, no?

mdouaihy: ok I understand now your point.

Unfortunately, weak symbols are not exposed by MSVC compiler. selectany works only on data items and not on methods.
https://docs.microsoft.com/en-us/cpp/cpp/selectany

However, we can use the */FORCE:MULTIPLE* link flag to force the link of the binary even if multiple definitions of a symbol exists
https://docs.microsoft.com/en-us/cpp/build/reference/force-force-file-output

Also, on windows, using dynamic linking, having the same symbols is not a problem.
This last point is important. If we take again the example, linking TracerA and TracerB statically with OT will cause duplication of the symbols of OT, among them the Global Tracer, which is not what we really want.

What do you think?

rnburn: Could you use selectany with a data variable that's a function pointer?


> If we take again the example, linking TracerA and TracerB statically with OT will cause duplication of the symbols of OT, among them the Global Tracer, which is not what we really want.

Static libraries are usually built as thin archives (i.e. without the dependent libraries linked in), so there wouldn't be duplication of the OT symbols since OT would be linked in by the app along with the tracing libraries.

mdouaihy: done.

mdouaihy:
@rnburn, changes done as per our discussion.

src/dynamic_load_unix.cpp
```diff
@@ -45,16 +45,16 @@ DynamicallyLoadTracingLibrary(const char* shared_library,
       new DynamicLibraryHandleUnix{handle}};
 
   const auto make_tracer_factory =
-      reinterpret_cast<decltype(OpenTracingMakeTracerFactory)*>(
+      reinterpret_cast<OpenTracingMakeTracerFactoryType**>(
           dlsym(handle, "OpenTracingMakeTracerFactory"));
-  if (make_tracer_factory == nullptr) {
+  if (make_tracer_factory == nullptr || *make_tracer_factory == nullptr) {
     error_message = dlerror();
```
rnburn: If `*make_tracer_factory == nullptr`, won't this need a different error message since `dlerror` wouldn't return anything?

mdouaihy: indeed. you're totally right.

include/opentracing/symbols.h
```diff
@@ -0,0 +1,40 @@
...
+// Export if this is our own source, otherwise import:
+#ifndef OPENTRACING_STATIC
+#ifdef OPENTRACING_EXPORTS
+#define OPENTRACING_API __declspec(dllexport)
+#else  // OPENTRACING_STATIC
+#define OPENTRACING_API __declspec(dllimport)
+#endif // OPENTRACING_EXPORTS
+#endif  // OPENTRACING_STATIC
+
+#define OPENTRACING_ATT
```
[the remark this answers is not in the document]
mdouaihy: yes. removed in the latest commit.

src/dynamic_load_windows.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  std::unique_ptr<DynamicLibraryHandle> dynamic_library_handle{
+      new DynamicLibraryHandleWindows{handle}};
+
+  const auto make_tracer_factory =
+      reinterpret_cast<OpenTracingMakeTracerFactoryType**>(
+          GetProcAddress(handle, "OpenTracingMakeTracerFactory"));
+
+  if (make_tracer_factory == nullptr) {
+    error_message = "An error occurred while looking up for OpenTracingMakeTracerFactory : " + GetLastError();
```
rnburn: This [StackOverflow question](https://stackoverflow.com/questions/1387064/how-to-get-the-error-message-from-the-error-code-returned-by-getlasterror) says that we need to use  `FormatMessage` along with `GetLastError` to get a useful message.

mdouaihy: Good idea. Change done. an example

```
Failed to load tracer library. An error occurred: The specified module could not be found.
```

rnburn:
Could we add some brief instructions to README.md for how to install on windows? The [current instructions](https://gitter.im/opentracing/opentracing-cpp?at=5ae8960ef2d2d53704552df9), I think, are unix/linux-specific.

mdouaihy:
sure. Very good idea.

rnburn:
@mdouaihy - Why'd you close?

mdouaihy:
I didnt do that on purpose. I might have pressed on the wrong button

mdouaihy:
Instructions to generate the Visual Studio solution, to run the build and the tests are added now. Both Release and Debug target are covered.

include/opentracing/dynamic_load.h
```diff
@@ -35,14 +36,45 @@
...
+  }  // extern "C"
+
+#else
+
+#define OPENTRACING_DECLARE_IMPL_FACTORY(X)                                 \
+  extern "C" {                                                              \
+                                                                            \
+  int __attribute((weak))                                                   \
+      X(const char* opentracing_version, const void** error_category,       \
+        void** tracer_factory);                                             \
```
rnburn: I don't think we need these declarations:

```cpp
+__declspec(dllexport) int X(const char* opentracing_version,                \
+                            const void** error_category,                    \
+                            void** tracer_factory);
```
```cpp
+  int __attribute((weak))                                                   \
+      X(const char* opentracing_version, const void** error_category,       \
+        void** tracer_factory); 
```
All that needs to be exposed is the variable `OpenTracingMakeTracerFactory`. `X` can be a static function in the translation unit.

mdouaihy: This declaration is useful so that 
```
OPENTRACING_DECLARE_IMPL_FACTORY(OpenTracingMakeTracerFactoryFct);
```
can be declared anywhere in the translation unit.

I'll do the necessary changes.

mocktracer/src/dynamic_load.cpp
```diff
@@ -4,7 +4,7 @@
 #include <cstring>
 #include <exception>
 
-int OpenTracingMakeTracerFactory(const char* opentracing_version,
+int OpenTracingMakeTracerFactoryFct(const char* opentracing_version,
```
rnburn: Can we make `OpenTracingMakeTracerFactoryFct` static?

mdouaihy: Sure.

rnburn:
@mdouaihy - This looks good. Was there anything else you wanted to add or should I go ahead and merge in?

mdouaihy:
@rnburn, nothing to add from my side for the moment. Once you merge, please create a version that we update on Hunter to finalize the Jaeger windows build.

rnburn:
@mdouaihy - I have another PR (https://github.com/opentracing/opentracing-cpp/pull/91) that also makes some changes to the dynamic loading API. I'll update it with the changes here, but I'd like to merge that before making the next release.

PR #96, opened 2018-06-20
Implements CPack for easy RPM building

gabrielkay:
This adds the option to generate two rpms for this repository, one rpm with the necessary .so files and one devel rpm with necessary headers. Calling `make rpm` after a general `make` generates the rpms into .build/BUILD/RPMS.


rnburn:
Could we use [CPackRPM](https://cmake.org/cmake/help/v3.7/module/CPackRPM.html) for this?

gabrielkay:
What would you want added through CPackRPM. In the end it calls rpm-build so it would be very similar. Also, I could work towards getting the rpms on packagecloud if this gets merged into the CI.

rnburn:
@gabrielkay - I prefer using CMake's builtin support when it's available.

One advantage of working off of CPack is that different packaging targets can share the same cpack configuration variables, requiring less work to maintain them.

And pushing releases to pagecloud through the CI definitely sounds good.

gabrielkay:
Take a look at the changes I just made. Is that what you were thinking? Let me know if I should change/add anything.

rnburn:
I don't believe you need those custom targets if you're using cpack.

Take a look at the RPM section here: https://gitlab.kitware.com/cmake/community/wikis/doc/cpack/PackageGenerators

gabrielkay:
I've never used CPack before and couldn't quite figure it out. I tried to make two components that would turn into two rpms but it is always making just one rpm with all files in it. If you could help me out by pointing out any CPack "sets" I'm missing or what exact command line I should run to generate the rpms that would be extremely helpful.

rnburn:
@gabrielkay - I moved the line
```
set(CPACK_RPM_COMPONENT_INSTALL ON)
```
to the top of the CMakeLists.txt file above
```
cmake_minimum_required(VERSION 3.1)
```
then ran with
```
/opentracing-cpp/build# cpack -G RPM
```
and it generated two RPMs
```
CPack: Create package using RPM
CPack: Install projects
CPack: - Run preinstall target for: opentracing-cpp
CPack: - Install project: opentracing-cpp
CPack: -   Install component: Devel
CPack: -   Install component: Unspecified
CPack: Create package
CPackRPM: Will use GENERATED spec file: /opentracing-cpp/build/_CPack_Packages/Linux/RPM/SPECS/opentracing-cpp-Devel.spec
CPackRPM: Will use GENERATED spec file: /opentracing-cpp/build/_CPack_Packages/Linux/RPM/SPECS/opentracing-cpp-Unspecified.spec
CPack: - package: /opentracing-cpp/build/opentracing-cpp-1.5.0-Linux-Devel.rpm generated.
CPack: - package: /opentracing-cpp/build/opentracing-cpp-1.5.0-Linux-Unspecified.rpm generated.
```

rnburn:
@gabrielkay - cpack should handle the files for you. If you take the take the existing opentracing-cpp repository, run 
```
git clone https://github.com/opentracing/opentracing-cpp.git
cd opentracing-cpp
mkdir build
cd build
cmake ..
cpack -G RPM --config CPackConfig.cmake
```
Then `rpm -qlp opentracing-cpp-1.5.0-Darwin.rpm` shows that all the headers and other files are included
```
/usr/include/opentracing
/usr/include/opentracing/config.h
/usr/include/opentracing/dynamic_load.h
...
```
What are you getting stuck with?

The [CMake tutorial](https://cmake.org/cmake-tutorial/) includes a section on cpack if you're looking for an example.

gabrielkay:
I agree that CPack works fine to just package everything into one RPM but I need to create two RPMs, one just with the libraries that I can link against (libopentracing.so.1.5.0 & libopentracing_mocktracer.so.1.5.0) and one for developing with all needed .h headers. I've been trying to use the component feature of CPack to do this but ran into issues when trying to create components and specify which files go into which component.

gabrielkay:
After a lot of trial and error, now CPack produces the two RPMs I need. They have the endings -DEVEL and -DIST. All that's needed to be done is calling `cpack -G RPM` in the .build directory after the directory is made. It also makes two more rpms by default (-Devel and no extension) but I see no way to prevent those two from being created. Looks to me like this is very close to being ready, let me know if you see something wrong.

CMakeLists.txt
```diff
@@ -228,3 +229,38 @@ endif()
...
+        DESTINATION include/opentracing/
+        COMPONENT DEVEL)
+install(FILES ${CMAKE_BINARY_DIR}/include/opentracing/config.h
+        DESTINATION include/opentracing/
+        COMPONENT DEVEL)
+install(FILES ${CMAKE_BINARY_DIR}/include/opentracing/version.h
+        DESTINATION include/opentracing/
+        COMPONENT DEVEL)
+install(FILES ${CMAKE_BINARY_DIR}/output/libopentracing_mocktracer.so.1.5.0
+        DESTINATION lib64/
```
rnburn: Could you use the LIB_INSTALL_DIR from https://github.com/opentracing/opentracing-cpp/pull/78 instead of hard coding lib64?

Also, we don't want to reference libraries like `${CMAKE_BINARY_DIR}/output/libopentracing_mocktracer.so.1.5.0` the so extension isn't portable and fixing the version like that will mean it breaks when we update.

Could you use the existing install commands but only with the COMPONENT added?

opentracing-cpp-devel.spec
```diff
@@ -0,0 +1,46 @@
+Name:           %{_name}
```
rnburn: Aren't these spec files generated by cpack?

CMakeLists.txt
gabrielkay: Yep, my bad for hard-coding the version number in right there. I'd been using the version variables for that for everything until I switched to CPack. And yes I'll switch the lib64.

opentracing-cpp-devel.spec
gabrielkay: That was my initial commit using those, I'll delete those.

gabrielkay:
Take a look at my changes if you have a chance. The rpms compile with `cpack -G RPM`. The only issue I'm encountering is I get the errors below when compiling with my project. Do you have any idea why these specific 3 calls break my compile? This error has been very frustrating.
`/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64/libzipkin_opentracing.so.0: undefined reference to opentracing::v1::propagation_error_category()`
`/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64/libzipkin_opentracing.so.0: undefined reference to opentracing::v1::dynamic_load_error_category()`
`/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64/libzipkin_opentracing.so.0: undefined reference to opentracing::v1::tracer_factory_error_category()`
`collect2: error: ld returned 1 exit status`

rnburn:
@gabrielkay - zipkin-opentracing hasn't been updated yet to the latest 1.5 release.

It's not linking because the ABI version number was bumped so that it has a different inline namespace.

If you want to add a PR to update it, it should look something like this: https://github.com/lightstep/lightstep-tracer-cpp/pull/105.

Otherwise, I'll be getting to it myself shortly.

CMakeLists.txt
```diff
@@ -228,3 +245,18 @@ endif()
...
+
+install(DIRECTORY ${CMAKE_SOURCE_DIR}/include/opentracing/
+        DESTINATION include/opentracing/
+        COMPONENT DEVEL)
+install(FILES ${CMAKE_BINARY_DIR}/output/libopentracing_mocktracer.so.${OPENTRACING_VERSION_STRING}
+        DESTINATION lib64/
+        COMPONENT DIST)
+install(FILES ${CMAKE_BINARY_DIR}/output/libopentracing.so.${OPENTRACING_VERSION_STRING}
+        DESTINATION lib64/
+        COMPONENT DIST)
```
rnburn: Instead these three installs, can the COMPONENT be added to the installs in the `if (BUILD_SHARED_LIBS)` block?

Also, the shared libraries won't be built if `-DBUILD_SHARED_LIBS=OFF` is specified.

ldconfig
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+ldconfig
```
rnburn: Could you move this to the cmake folder and add a short note for what it's used for?

CMakeLists.txt
gabrielkay: Installing through the existing BUILD_SHARED_LIBS call would include multiple unnecessary .so files (libopentracing.so, libopentracing.so.1, libopentracing.so.1.5.0) when only the last one is wanted and needed.

gabrielkay:
Now the DIST rpm is only made when BUILD_SHARED_LIBS is on. The script was clarified, renamed, and moved.

CMakeLists.txt
```diff
@@ -154,9 +164,17 @@ if (BUILD_SHARED_LIBS)
                                              SOVERSION ${OPENTRACING_VERSION_MAJOR})
   target_compile_definitions(opentracing PRIVATE OPENTRACING_EXPORTS)
   install(TARGETS opentracing EXPORT OpenTracingTargets
-          RUNTIME DESTINATION ${LIB_INSTALL_DIR}
-          LIBRARY DESTINATION ${LIB_INSTALL_DIR}
-          ARCHIVE DESTINATION ${LIB_INSTALL_DIR})
+          RUNTIME DESTINATION ${LIB_INSTALL_DIR}64
+          LIBRARY DESTINATION ${LIB_INSTALL_DIR}64
+          ARCHIVE DESTINATION ${LIB_INSTALL_DIR}64
```
rnburn: The idea with LIB_INSTALL_DIR is that it's set from the command line to override the default location when you invoke cmake, e.g.,
```
cmake -DLIB_INSTALL_DIR=lib64 ..
```
So could you remove the `64` suffixes and use `-DLIB_INSTALL_DIR=lib64` if you want to build the package.

CMakeLists.txt
```diff
@@ -154,9 +164,17 @@ if (BUILD_SHARED_LIBS)
...
+          )
+  install(FILES ${CMAKE_BINARY_DIR}/output/libopentracing_mocktracer.so.${OPENTRACING_VERSION_STRING}
```
rnburn: Using `.so` like that is going to be a problem on OSX and Windows.

Could you add a `make install` command [here](https://github.com/opentracing/opentracing-cpp/blob/master/.travis.yml#L50) so that we have test coverage of the installation? and then fix the OSX error.

gabrielkay: I made the two changes suggested in the comments above but could you clarify what you mean here.

rnburn: OSX doesn't use `.so` for shared libraries, when you run `make install` it gives this error
```
-- Installing: /usr/local/lib64/libopentracing.dylib
CMake Error at cmake_install.cmake:55 (file):
  file INSTALL cannot find
  "/tmp/opentracing-cpp/build/output/libopentracing_mocktracer.so.1.5.0".


make: *** [install] Error 1
```

gabrielkay:
I don't use OSX, how can that be fixed? Can CMake check the system's architecture?

rnburn:
Maybe use [CMAKE_SHARED_LIBRARY_SUFFIX](https://cmake.org/cmake/help/v3.0/variable/CMAKE_SHARED_LIBRARY_SUFFIX.html)

gabrielkay:
CMAKE_SHARED_LIBRARY_SUFFIX worked fine for me on Linux. I added make install to the travis.yml so we can see if it fixed the OSX issue.

rnburn:
@gabrielkay - what happened to you separate installs for the version shared library? Did you end up not needing it?

gabrielkay:
OSX was using a naming convention of libopentracing.1.5.0.dylib with number before .dylib while the .so files had number after .so so using the variable didn't work. I just put made component a part of the shared library install for opentracing and installed the unnecessary .so's as well.  I also stopped packaging the libopentracing_mocktracer because there is no install call for it in the CMake so I couldn't package it without specifying its name. How is the libopentracing_mocktracer installed? I would prefer to package it here but I might just have to add the call back to my own branch if it breaks OSX

rnburn:
The mocktracer has its own cmake file, but you should be able to add the components to the installs in [here](https://github.com/opentracing/opentracing-cpp/blob/master/mocktracer/CMakeLists.txt)

gabrielkay:
That works perfectly!

gabrielkay:
I think this is finally ready to be merged?

mocktracer/CMakeLists.txt
```diff
@@ -18,7 +18,9 @@ if (BUILD_SHARED_LIBS)
                                                SOVERSION ${OPENTRACING_VERSION_MAJOR})
   target_link_libraries(opentracing_mocktracer PUBLIC opentracing)
   target_compile_definitions(opentracing_mocktracer PRIVATE OPENTRACING_MOCK_TRACER_EXPORTS)
-  install(TARGETS opentracing_mocktracer EXPORT OpenTracingTargets
+  install(TARGETS opentracing_mocktracer
+          COMPONENT DIST
+          EXPORT OpenTracingTargets
```
rnburn: Should the headers for the mocktracer (below) be installed into the DEVEL component?

gabrielkay: I had no use for them. They could be installed too but it would have to be in a subfolder because some share names that would overwrite other files.

rnburn: You can still use the mocktracer without them, but only from the dynamic loading interface.

All the headers should already be in the mocktracer folder so they shouldn't clash.

Up to you whether you want to install them or not, I'm ok either way.

gabrielkay: I'm fine without them.

PR #101, opened 2018-12-11
Add IsGlobalTracerRegistered to Tracer

MikeGoldsmith:
Adds support for knowing if a global tracer has been registered via a global variable including tests.

Issue first reported here: https://github.com/opentracing/opentracing-python/pull/109

This change has been accepted into C#, Java and Go and there are pending reviews for Python and NodeJS.

### DIFF
--- a/src/tracer.cpp
+++ b/src/tracer.cpp
@@ -8,6 +8,11 @@ static std::shared_ptr<Tracer>& get_global_tracer() {
   return global_tracer;
 }
 
+static std::atomic<bool>& get_is_global_tracer_registered() {
+  static std::atomic<bool> is_global_tracer_registered{false};;
+  return is_global_tracer_registered;
+}
+
 std::shared_ptr<Tracer> Tracer::Global() noexcept {
   return get_global_tracer();
 }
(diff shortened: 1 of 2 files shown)


MikeGoldsmith:
I believe Tracer::Global would return an instance of a NoopTracer by default, or the registered instance if SetTracer was called.

The IsGlobalTracerRegistered function needs to return false until a tracer has intentionally been registered.

src/tracer.cpp
```diff
@@ -8,14 +8,24 @@ static std::shared_ptr<Tracer>& get_global_tracer() {
...
+}
+
 std::shared_ptr<Tracer> Tracer::Global() noexcept {
   return get_global_tracer();
 }
 
 std::shared_ptr<Tracer> Tracer::InitGlobal(
     std::shared_ptr<Tracer> tracer) noexcept {
   get_global_tracer().swap(tracer);
+  get_is_global_tracer_registered().store(true);
```
rnburn: There's a race condition here if IsGlobalTracerRegistered gets called between lines 22 and 23.

MikeGoldsmith:
@rnburn - I've added a mutex to protect against the race condition when setting the global tracer. Let me know if you would prefer a alternate solution.

I think this is an edge case as `InitGlobal` is likely to only be called during process startup but happy to fix as above.

rnburn:
Implemented in #116

PR #103, opened 2019-02-08
Overload convert_time_point for equal clocks

Oberon00:
This is faster and more precise if `convert_time_point` is called with equal ToClock and FromClock.

### DIFF
--- a/include/opentracing/util.h
+++ b/include/opentracing/util.h
@@ -36,6 +36,13 @@ class option_wrapper {
   const T *ptr_;
 };
 
+template <class SameClock, class Duration>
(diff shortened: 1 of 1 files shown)


Oberon00:
As you can see from my hasty second commit, this is rather untested (it compiled on MSVC, but seems currently unused and thus uninstantiated)

Oberon00:
Actually, this implemenation does not work because the overloads are ambiguous if the new one can be instantiated, this would probably need tag dispatching or SFINAE. So take this PR as a suggestion. 😃 If there is interest, I can update the PR again to fix this.

rnburn:
Implemented in #115

PR #104, opened 2019-02-12
Fix MSVC compiler flags.

Oberon00:
* All debug specific flags would be replaced by release specific on MSVC.
* The OPENTRACING_STATIC flag would be missing from OpenTracingConfig.cmake when linking against OpenTracing::opentracing-static

I'd recommend also adding -D_CRT_SECURE_NO_WARNINGS and /W4 flags but wanted to keep this PR minimal. The changes here allow to build opentracing-cpp in debug mode on MSVC and link against it from another CMake project.


rnburn:
👍

Issue #105, opened 2019-04-08
OpenTracing in Vcpkg


david-antiteum:
More than an issue an announcement :)

I have added OpenTracing to vcpkg (C++ Library Manager) repository via this merged PR: https://github.com/Microsoft/vcpkg/pull/5974

It works in Windows and Linux (probably in macOS but vcpkg doesn't run in macOS yet).

During the porting I have found two problems:
- no UWP support (not important)
- cannot build the static version without the dynamic one (so vcpkg builds either the dynamic version or the static AND dynamic version)

Install the package with:

```
vcpkg install opentracing
```

Then use it in cmake (dynamic version):

```
find_package(OpenTracing CONFIG REQUIRED)
target_link_libraries(main PRIVATE OpenTracing::opentracing)
```

or (static version):

```
find_package(OpenTracing CONFIG REQUIRED)
target_link_libraries(main PRIVATE OpenTracing::opentracing-static)
```

rnburn:
Did you try building with `cmake -DBUILD_SHARED_LIBS=OFF`? -- that should produce only the static library.

david-antiteum:
Yes, I tried that.

Using this command:

```
vcpkg install opentracing --triplet x64-windows-static
```

That does this definitions in a modified portfile

```
-DBUILD_SHARED_LIBS=OFF
-DBUILD_STATIC_LIBS=ON
-DBUILD_MOCKTRACER=OFF
-DBUILD_DYNAMIC_LOADING=OFF
```

I got this errors (same with BUILD_DYNAMIC_LOADING=ON):

```
tracer_test.cpp.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) class std::error_category const & __cdecl opentracing::v2::propagation_error_category(void)" (__imp_?propagation_error_category@v2@opentracing@@YAAEBVerror_category@std@@XZ) referenced in function "void __cdecl opentracing::v2::`dynamic initializer for 'invalid_carrier_error''(void)" (??__Einvalid_carrier_error@v2@opentracing@@YAXXZ)
[...]

PR #106, opened 2019-04-22
compile mocktracer test using static lib #105

david-antiteum:
Adding missing define OPENTRACING_MOCK_TRACER_STATIC


Oberon00:
If I understand correctly, everyone that links against `opentracing_mocktracer-static` must define `OPENTRACING_MOCK_TRACER_STATIC`. If that is correct, I would suggest simply changing https://github.com/opentracing/opentracing-cpp/blob/32a0d7ba196bfbda4c29d00897f723c64308f5ca/mocktracer/CMakeLists.txt#L36 to make the definition PUBLIC instead of PRIVATE, so it will be automatically propagated.

david-antiteum:
Yes, you are right. Change done.

PR #108, opened 2019-05-31
Directly support opentracing::string_view as a Value.

rnburn:
This avoids unnecessary memory allocations and is similar to how `const char*` is already handled.


Oberon00:
Note that this is an ABI breaking change.

rnburn:
Yep, there's already this ABI breaking change that went in https://github.com/opentracing/opentracing-cpp/pull/56 so I'm trying to bundle up all the other ABI-breaking changes that have been pending for the next release so that I don't have to break it again in the near future.

PR #110, opened 2019-06-28
Support Building with Bazel on Windows

isaaczinda:


BUILD.bazel
```diff
@@ -1,12 +1,44 @@
...
+  includes=["include"],
+  deps = [
+    "//3rd_party:expected",
+    "//3rd_party:variant",
+  ],
+  # sets the OPENTRACING_EXPORTS precompiler variable while building DLL
+  # so that header "import symbols" tags become "export symbols" tags
+  copts=["/DOPENTRACING_EXPORTS"],
+)
+
```
rnburn: Would a windows user be able to also link statically to opentracing-cpp if they wanted to?

bazel/cross_platform.bzl
```diff
@@ -0,0 +1,20 @@
+def cc_posix_binary(name, srcs = [], **kargs):
```
isaaczinda: @rnburn the CI is failing because [the way that envoy does distro-specific libraries](https://github.com/envoyproxy/envoy/blob/master/bazel/envoy_library.bzl#L74-L101) doesn't translate to building binaries. Since the cc_binary rules is expected to produce outputs, `bazel build //...` still fails because, for example, `cc_windows_binary` produces no outputs on a mac. 

I'm looking for more advice regarding how to solve this problem.

BUILD.bazel
```diff
@@ -1,51 +1,81 @@
...
-    linkopts = [
-      "-ldl",
-    ],
+    ] + select({
+        "@bazel_tools//src/conditions:windows": [":opentracing_dll_import"],
+        "//conditions:default": [":opentracing_so_import"],
+    }),
+    # linkopts = [
+    #   "-ldl",
+    # ],
```
isaaczinda: @rnburn I'm not sure what the purpose of this linker flag is and haven't been able to find docs for it online.

BUILD.bazel
isaaczinda: For now, I'm going to hold off on supporting static linking because I think it will be a challenge. I was having a bunch of issues with this, and then found [this thread](https://github.com/bazelbuild/bazel/issues/1920) which explains how cc_binary isn't setup to build static libraries.

BUILD.bazel
```diff
@@ -1,19 +1,82 @@
...
     deps = [
       "//3rd_party:expected",
       "//3rd_party:variant",
-    ],
-    linkopts = [
-      "-ldl",
-    ],
+    ] + select({
+        "@bazel_tools//src/conditions:windows": [":opentracing_dll_import"],
+        "//conditions:default": [":opentracing_so_import"],
```
rnburn: Let's keep the opentracing as a static library and have the shared objects as separate targets.

BUILD.bazel
rnburn: dl is the library for the dlopen and related system calls. See http://man7.org/linux/man-pages/man3/dlopen.3.html

BUILD.bazel
rnburn: No, we need opentracing to be a static library. There are existing builds that depend on that and will break if it's changed.

A thin archive is the default output for `cc_library`. That thread you link to is something different, a thick archive with all of the dependencies linked in -- that's not necessary. 

Just keep cc_library the same but select the correct dynamic_load source for windows.

PR #128, opened 2020-03-04
Fix comment for Tracer::StartSpan

alnr:
Beside the comment above `class SetTag`, the comment changed in this PR is the only documentation on how to tag a span at creation time.

### DIFF
--- a/include/opentracing/tracer.h
+++ b/include/opentracing/tracer.h
@@ -84,7 +84,7 @@ class OPENTRACING_API Tracer {
   //     auto span = tracer.StartSpan(
   //         "GetFeed",
   //         {opentracing::ChildOf(&parentSpan.context()),
-  //         opentracing::Tag{"user_agent", loggedReq.UserAgent},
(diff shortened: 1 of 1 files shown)


include/opentracing/tracer.h
```diff
@@ -84,7 +84,7 @@ class OPENTRACING_API Tracer {
   //     auto span = tracer.StartSpan(
   //         "GetFeed",
   //         {opentracing::ChildOf(&parentSpan.context()),
-  //         opentracing::Tag{"user_agent", loggedReq.UserAgent},
+  //         opentracing::SetTag{"user_agent", loggedReq.UserAgent},
```
yurishkuro: I think the intention of this comment was actually correct, it's the class that's misnamed. Changing class name is a breaking change, but could we deprecated it and create a new one called `Tag`?

alnr: Or we could simplify everyone's life massively by living with the slightly misnamed class ;)

PR #138, opened 2021-07-21
Update catch2 LICENSE.txt to be a text file

### LINKED ISSUE #137: Broken error page instead of license for catch2
Hi! I noticed that the license for catch2 is not the license but is instead a github error page: https://github.com/opentracing/opentracing-cpp/blob/master/3rd_party/include/opentracing/catch2/LICENSE.txt. Can this please be replaced with an actual license? 😄

ssgelm:
I noticed that the LICENSE.txt for catch2 was a copy of the github html for the file, not just the file itself. This fixes that. 

Resolves #137


yurishkuro (approved):
Source https://github.com/catchorg/Catch2

PR #143, opened 2022-01-13
fix: use std::invoke_result instead of std::result_of

trezz:
std::result_of was deprecated in c++17 and removed in c++20.

This PR configures the project to use c++17 if the compiler supports it and use std::invoke_result instead of std::result_of.

If the compiler doesn't support c++17, the project is configured to fallback to c++11 and still use std::result_of.


perseoGI:
Any update on this????

perseoGI:
Is anyone taking care of this library?

